PR #12, opened 2017-07-15
Render GuideSidebar below content on mobile viewports

aweary:
Right now the mobile navigation experience is kind of painful, especially for the guide. There's no indication of what you should do when you hit the guide landing page.

This change makes it so the `GuideSidebar` component is always rendered and available on all viewports, without having to toggle a menu. The `GuideSidebar` component no longer needs to be toggleable, as it is just rendered below the content on mobile viewports and to the side of the content on desktop viewports. This makes it much easier to navigate through the guide on mobile.

This is similar to what React does https://facebook.github.io/react

# Before

screen shot 2017-07-15 at 11 04 09 am

# After 

screen shot 2017-07-15 at 11 04 17 am


aweary:
You can preview the change at http://mobile-nav-reasonml-docs.surge.sh/

jaredly:
I love the idea, btw. I think it'll be a great improvement to the mobile experience

PR #13, opened 2017-07-15
Use icon for "Suggest an Edit" link

aweary:
The "Suggest an edit" text felt a little verbose and was pushing the section title to a new line on mobile. This change uses a simple edit icon instead, taken from https://material.io/icons/

You can preview the change here: http://suggest-edit-icon-reasonml.surge.sh/

# Before

screen shot 2017-07-15 at 11 31 45 am

# After

screen shot 2017-07-15 at 11 31 31 am


chenglou:
@jaredly what do you think? For desktop I feel the word is more explicit. Especially now that the github link points to the actual reason repo rather than this one.

aweary:
@chenglou we could render the icon on mobile and the text on desktop

chenglou:
I think text on desktop and icon on mobile sounds nice?

jaredly:
Yeah! Let's use a media query to swap between the two

aweary:
@chenglou @jaredly done! I've updated http://suggest-edit-icon-reasonml.surge.sh with the change as well.

chenglou:
Looking good! Thanks =)

PR #23, opened 2017-07-17
Add generated API docs to the website

jsdf:
I generated some stdlib API docs for the site by doing some transformations on the HTML output of ocamldoc. The process of generating the docs is quite heavy so I think it makes sense to just check in the HTML.

screen shot 2017-07-16 at 11 17 34 pm


aweary:
Is there any significant advantage to duplicating this in the ReasonML docs over just linking to the existing OCaml docs and recommending reason-tools?

jsdf:
@aweary I would say that having to go through that process is a source of confusion and an unnecessary hurdle to new users.

Having this stuff on the Reason site may also help when people google something like "reason array".

aweary:
I can appreciate that, it does side-step that process and help SEO. My biggest reservation is honestly that these existing OCaml docs just...aren't that approachable. I still have a hard time reading through them and I think their reliance on type signatures over examples make them less accessible to newer users--especially in the context of Reason, where we're heavily targetting JavaScript developers.

I'm worried it might scare some people away. I'd be great if we could provide a set of better API docs though.

jsdf:
I absolutely agree about the existing OCaml docs being unfriendly and I only see this as a stopgap until we can produce better docs (and ideally a more friendly, approachable stdlib). However, for the time being until we have that, would bumping people over to the OCaml docs be any less scary?

aweary:
> Perhaps we could add some content to the index page and common modules (eg. List, Array) which directs people to the relevant guides first, and explains the situation with the docs?

I think that's a great idea. If we can frame the situation beforehand ("we know they're unfriendly, but in case you want to read them...") it'll be less likely to deter new users.

chenglou:
Speaking as someone who has no trouble navigating the ocaml api docs: people _keep_ asking in the past year where they can find documentations on the stdlib. So I don't mind at all us inlining the docs.

Thanks for styling it, too!

chenglou:
@jaredly what do you think of this approach?

aweary:
@chenglou it's worth recognizing that you (and those who have worked with Reason in the last year)  represent a subset of people who may be far more comfortable reading OCaml documentation.

I think we can do better than inlining, such as what @jsdf mentioned with adding introductory content, or even examples. I think that's in line with Reason's goal of making OCaml more accessible.

chenglou:
Yeah I was speaking _for_ this effort =)

aweary:
Right! And what I'm saying is that the effort of inlining the OCaml API documentation with the Reason syntax could be a problem for new users who _aren't_ comfortable reading OCaml documentation. 

Nobody wants to click an API page and immediately get a page full of type signatures they don't understand. That's scary. What I'm advocating is adopting these docs but with qualifications; we add introductory content or even examples so we can soften the blow, so to speak.

chenglou:
Understood. I think we should merge this and incrementally improve this; people who need to read stdlib docs will have to read them anyway, and gently discouraging that (Is that the right term?) through an extra link isn't going to change much. In that sense this PR is s step better, at least.

Right?

kennetpostigo:
2 cents here: I had started porting over the docs with human friendly taste and styles:
https://twitter.com/kennetpostigo/status/840757506716622848

and the actual markdown can be found here: https://github.com/kennetpostigo/rdn/tree/master/documentation/native/reason

I think I have more complete documentation that I wrote but haven't pushed up. I just need to find my old laptop.

chenglou:
Discussed in discord. Merged!

Thanks so much for the work. Let's keep on improving this!

PR #24, opened 2017-07-17
Add editor plugin atom-ide-reason

zaaack:
My nuclide's language support was broken yesterday, but I tried [atom-languageclient](https://github.com/atom/atom-languageclient) and surprising discovered atom-langugageclient can work perfectly with the master branch of [ocaml-language-server](https://github.com/freebroccolo/ocaml-language-server) (plus the only un-merging PR). 

So I create a package [ide-reason](https://github.com/zaaack/atom-ide-reason) and hoping it could help more people.

### DIFF
--- a/src/pages/guide/tools.md
+++ b/src/pages/guide/tools.md
@@ -108,8 +108,8 @@ https://github.com/freebroccolo/vscode-reasonml
 ### Atom
 
 You can install the [Atom](https://atom.io/) Reason integration through either
-[atom-ocaml-merlin](https://atom.io/packages/ocaml-merlin) or
-[Nuclide](https://nuclide.io/). The former is lighter.
+[atom-ocaml-merlin](https://atom.io/packages/ocaml-merlin), [atom-ide-reason](https://github.com/zaaack/atom-ide-reason) or
+[Nuclide](https://nuclide.io/). The first two is lighter.
 
 <img style="width:100%; max-width:470px; max-height:440px" src="images/AtomAutocomplete.png" />
 
@@ -124,6 +124,12 @@ As noted on the page, you'll also need the linter (which itself needs Linter-ui-
 Intention and Busy-signal).
 
 **Note**: if you've installed the global binaries through the [JS Workflow](./gettingStarted.html#javascript-workflow-project-setup) then you can ignore the `opam install merlin` instructions at the end.
+
+#### Atom-IDE-Reason
+An atom-languageclient plugin using the same language-server of vscode-reason, as the [readme](https://github.com/zaaack/atom-ide-reason) says, you'll also need:
+
+- [atom-ide-ui](https://atom.io/packages/atom-ide-ui)
+- [language-reason](https://atom.io/packages/language-reason)
 
 #### Nuclide
 Alternatively, if you use Nuclide:


chenglou:
Oh wow, cc @jsdf
If this works then we should switch out the piecemeal atom packages. Does this obsolete them? e.g. reason-refmt?

zaaack:
@chenglou Currently I only combined languageclient & lanuguage server, which provide linter, autocomplete, typehint, but still need language-reason for syntax hightlight, and reason-refmt for code-format. There might be some bug with reason-refmt's linter option, I'm using community's linter package(seems atom-ide-ui's diagnostics-ui is a little out-of-date) and have to disable it.

chenglou:
Thanks to this PR I've finally gotten around looking into language server and clients. They look really nice!

Though I'm weary of adding yet another option to our atom recommendations; ideally it's just this, fronted by either atom-languageclient or nuclide. Is there a reason why syntax highlight and reason-refmt aren't rolled back into the client? If I'm not mistaken, they're implemented in ocaml-language-server already.

zaaack:
I checked ocaml-language-server and vscode-reason again, code formating is a built-in feature in ocaml-language-server(I'll remove it), but syntax highlight is still needed, look here, vscode-reason implement it inside the package: https://github.com/reasonml-editor/vscode-reasonml/tree/master/src/syntaxes

jaredly:
I think it makes sense (or at least follows the conventions) for syntax highlighting to be separate

chenglou:
I've just tried this, every feature seems to work =)

There are however two blockers/bugs:

1. Opening atom through GUI (aka not terminal, and not at the root) doesn't trigger merlin still.
2. There are some weird (concurrency maybe?) issues regarding false-positive merlin reports. Not the usual, occasional merlin false positives. Here's a repo:

```reason
let foo bar baz => 1;

let a =
  foo
    "hiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii"
    (
      fun a =>
        ignore "byyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyye"
    );
```

Open this file, put an empty line somewhere, then format code. There'll be a new merlin error at the last line (expecting a structure item).

zaaack:
@chenglou I'll look into this.

chenglou:
Thanks! Fwiw I'm very looking forward to this. This'll suddenly unlock lots of editor integrations goodies for us!

chenglou:
Confirmed with an Oni user that this appears too. So an ocaml-language-server issue. If we fix it we fix for all editors

chenglou:
But then again, never had this happen in vscode proper

zaaack:
@chenglou The second one I can reproduce, but the first one I can't, could you please tell me your OS and environment? I have test in  MacOS and linux, but my reason was installed by opam, and OCaml version is 4.02.3+buckle-master(via bucklescript guid)

zaaack:
The second one is a little tricky and I gave up of digging, so I create an issue here: https://github.com/atom/atom-languageclient/issues/72

arxanas:
Not sure which of you are also Facebook employees (waleedk here, you can ping me on Messenger), but you should know that I've been investing some time in getting `ocaml-language-server` integrated with Nuclide.

On the Hack team, we've been pushing to implement the LSP in Nuclide so that we can use Hack over it. A side effect is that I can probably significantly improve the Hack team's development workflow if I can get OCaml support also using the LSP.

Let me know if there's anything you want to coordinate about getting Reason supported by the same mechanisms. (Maybe cc @jordwalke, I don't know if he knows that I work at Facebook.)

chenglou:
I'm at Facebook too. Let's talk!

chenglou:
@zaaack I've rewritten the editor tooling section and added your plugin. It'd be great to have it stable, so that we can unify around a single solution.

Symbolically merged this diff =) Pretty excited for ocaml-language-server

PR #28, opened 2017-07-17
Move markdown comment outside of verbatim block

kini:
The markdown hyperlink inside the code comment wasn't rendering into an actual hyperlink, so I moved it outside of the verbatim block into a body paragraph.


chenglou:
Thanks!

PR #32, opened 2017-07-18
Minor fixes in JSX section

arnarthor:
Fixed a minor issue with the JSX transformation where components are not called using `createElement` but `make`.
 Reference [here](https://github.com/facebook/reason/blob/1db2f97762b210b3be54439f3a33bd0bde2b65c2/src/reactjs_jsx_ppx.ml#L89)

Also changed an issue with text as children. They need to be converted to reactElements using `ReasonReact.stringToElement` the previous example would not compile.

### DIFF
--- a/src/pages/guide/language/jsx.md
+++ b/src/pages/guide/language/jsx.md
@@ -14,7 +14,7 @@ Capitalized tag:
 becomes
 
 ```reason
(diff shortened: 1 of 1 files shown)


jaredly:
awesome, thanks!

PR #34, opened 2017-07-19
fix rendering of API docs 'type' pages

jsdf:
The whitespace on pages like https://reasonml.github.io/api/type_Parsetree.html is not being preserved correctly. Fixed in this PR.

screen shot 2017-07-18 at 10 44 37 pm


jaredly:
coolio

PR #35, opened 2017-07-19
Minor changes in data types section

MoOx:
I was confused because I read `let horsePower` (my eyes/brain omitted the 1) and so I was confused by the type car below because I was seing a tuple used in the type. Using a more different name for the tuple and the type should avoid this for someone else!


chenglou:
Thanks! Keep these coming =)

PR #38, opened 2017-07-19
Update ocaml.md

MoOx:
Fix some markdown not correctly parsed (title are "broken" on the website).
Maybe we should lint and/or autofix this kind of issues :)


chenglou:
Thanks!

PR #41, opened 2017-07-20
Update syntax-cheetsheet.md

Gregoirevda:
I found it sometimes unclear what code block belongs to what description. This happens every time a section (###) has multiple descriptions + code blocks. I Added a subtitle every time that happens, to separate each description + code block


chenglou:
Thanks!

Issue #9, opened 2017-07-15
Next page at the bottom of each section


arnarthor:
I think it makes sense to add a next page button in the bottom right corner of each chapter/section. It gets tedious to scroll up (on small screens) after reading a section to find the next one.

chenglou:
Done!

PR #42, opened 2017-07-20
Add 'Previous' and 'Next' links to guide pages

jsdf:
The React docs have this feature and I think it's quite good for helping new users find their way through the guide.

screen shot 2017-07-19 at 9 30 32 pm


chenglou:
Fixes #9

Thanks, as always

PR #45, opened 2017-07-20
Fix return type

albertodotcom:
using `void` instead of `unit` I get `Unbound type constructor void`


jaredly:
thanks!

albertodotcom:
No problem, you are doing a great job!!!

PR #57, opened 2017-07-28
bug(QuickStart): Change link to be absolute path

kyleholzinger:
On the website, it will link to
`https://reasonml.github.io/https://reasonml.github.io/reason-react/gettingStarted.html`
instead of the intended
`https://reasonml.github.io/reason-react/gettingStarted.html`

### DIFF
--- a/src/pages/guide/javascript/quickstart.md
+++ b/src/pages/guide/javascript/quickstart.md
@@ -18,6 +18,6 @@ npm run watch
 ```
 This will compile Reason to Javascript in the `lib/js/` folder
 
-Alternatively, **to start a [ReasonReact](https://reasonml.github.io/reason-react/gettingStarted.html) app**, try `bsb -init my-react-app -theme react`.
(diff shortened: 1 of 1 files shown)


kyleholzinger:
Not sure if the site reads from this markdown file, but if you click on that link currently it's broken I think because of this

chenglou:
Aka good catch. That might be a bug in the static site generator. When the link starts with the same domain (https://reasonml.github.io) it's probably catching that and preventing a refresh or something, even though it's a different site. Doesn't happen locally when the url is localhost.

Thanks!

cc @KyleAMathews

KyleAMathews:
Normally you want to use the relative path e.g. for this link `/reason-react/gettingStarted.html`

KyleAMathews:
Oh I see, these are different sites — then yeah, you need to use an absolute path as otherwise it does catch the link to try to do a local redirect.

We could support "protocol-relative URLs" and not catch those but since https is standard these days, [using protocol-relative URLs is an anti-pattern](https://www.paulirish.com/2010/the-protocol-relative-url/)

PR #60, opened 2017-07-29
Delete % char

Gregoirevda:
Shouldn't it only be 1 `%`? Other examples only show 1 `%` and there's no explanation on 2 or 1


Gregoirevda:
Closing: 

The difference between the 2 %% from the previous section and the 1 % here is important! [%%something ...] is an OCaml "extension point" that represents a top-level statement (it can't show up inside a function or value, for example). [%something ...] is an extension point that stands in for an expression, and can be put just about anywhere -- but make sure that the JavaScript you put inside is actually an expression! E.g. don't put a semicolon after it, or you'll get a syntax error when you try to run the resulting JavaScript.

PR #62, opened 2017-07-30
Add documentation on record interop with javascript

Gregoirevda:
When  using a js function from an external package that is expecting an object as param, you can't pass him a Record.
Explain how to handle that


chenglou:
Thanks! I've tweaked the example a bit.

Gregoirevda:
Great!
So { "prop": "value" } === [%bs.obj { prop: value }] 

btw, I sometimes see  [%bs.obj { prop: value }]  and sometimes  [%bs.obj { prop= value }] 
(: vs =) what's the difference?

chenglou:
The latter isn't valid. You're probably thinking of OCaml syntax?

Gregoirevda:
Yes, I saw the latter one in Bucklescript documentation. Ok, so it's reasonML vs OCaml syntax. Though it might be assignment vs type someting like that

PR #69, opened 2017-08-04
Sharable Try links

chidelmun:
This PR implements Sharable try links.

### DIFF
--- a/src/pages/try.js
+++ b/src/pages/try.js
@@ -30,7 +30,19 @@ const bs = text => {
   }
 }
 
-const defaultSource = 'let x = 10;\nJs.log x;'
+const decodeBase64 = () => {
+    let url_encode = window.location.href;
+    let url = new URL(url_encode);
+    let digestCode = url.searchParams.get("digest");
+    if (digestCode == null) {
+      return 'let x = 10;\nJs.log x;'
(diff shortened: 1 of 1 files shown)


src/pages/try.js
```diff
@@ -57,7 +69,23 @@ export default class Try extends Component {
     jsIsLatest: false,
     autoEvaluate: true,
     output: [],
+    tryUrl: window.location.href,
+  }
+
+  handleClick = this.handleClick.bind(this);
+
+   handleClick() {
```
jaredly: you can use a class attribute to autobind it, e.g.

```
handleClick = () => {
```

src/pages/try.js
```diff
@@ -57,7 +69,23 @@ export default class Try extends Component {
...
+    var val = this.encodeBase64(this.state.reason);
+    this.setState({tryUrl: val});
+    document.getElementById('digest').disabled = false;
```
jaredly: using IDs & getElementById is super not reacty :/
instead of having that input, could we just do `window.location = val`? Then ppl can copy/paste the url from their browser

chidelmun:
Thanks. @jaredly. I will  address those immediately.

PR #70, opened 2017-08-04
Fixed PR.

chidelmun:
@jaredly I have addressed the changes you mentioned. Please review this.


chenglou:
Thanks! The Try link breaks on a second click from this diff

chidelmun:
Oh I see why. Will fix it right away. Thanks @chenglou

src/pages/try.js
```diff
@@ -57,7 +69,21 @@ export default class Try extends Component {
     jsIsLatest: false,
     autoEvaluate: true,
     output: [],
+    tryUrl: window.location.href,
```
jaredly: we no longer need `tryUrl` to live on state I think

src/pages/try.js
```diff
@@ -57,7 +69,21 @@ export default class Try extends Component {
...
+
+   handleClick = ()=> {
+    var val = this.encodeBase64(this.state.reason);
+    this.setState({tryUrl: val});
+    window.location = val;
+  }
+
+  encodeBase64(input){
+        this.setState({tryUrl : ""});
+        return  window.location.href +"?digest=" + window.btoa(input);
```
jaredly: nit: indent looks off

src/pages/try.js
chidelmun: Yes.  I cleaned it out. Thanks for pointing that.

jaredly:
looks like a `.swp` file crept in?

src/pages/try.js
```diff
@@ -30,7 +30,19 @@ const bs = text => {
   }
 }
 
-const defaultSource = 'let x = 10;\nJs.log x;'
+const decodeBase64 = () => {
+    let url_encode = window.location.href;
+    let url = new URL(url_encode);
+    let digestCode = url.searchParams.get("digest");
```
chenglou: This isn't available on IE unfortunately; also, the getter will return an already encoded snippet, so when you persist "foo bar" into the URL, here you'd get back "foo+bar". Which is then decoded into an invalid snippet.

src/pages/try.js
```diff
@@ -58,6 +70,17 @@ export default class Try extends Component {
     autoEvaluate: true,
     output: [],
   }
+
+  handleClick = ()=> {
+    var val = this.encodeBase64(this.state.reason);
+    window.location = val;
```
chenglou: This will cause an unnecessary refresh of the page. We should use `window.history.replaceState`

src/pages/try.js
```diff
@@ -217,6 +240,9 @@ export default class Try extends Component {
         <div css={{ backgroundColor: accent, color: 'white' }}>
           <Header inverted />
         </div>
+        <center>
+          <span> <button onClick={this.handleClick} css={styles.ButtonStyle}>Create Tryit Link</button> </span>
```
chenglou: I'm thinking we should just debounce and then serialize the code into url every 100ms. Might be potentially distracting; let's see if that works though

chenglou:
Alright, time to merge this! I'm gonna apply the changes from my comments subsequently. Thanks for the PR! =D

PR #75, opened 2017-08-07
Simplifies MakeSet functor using pointfree style

mikaelbr:
Removes unnecessary point from Set functor example. Minor trivial simplification of the code.


chenglou:
Humm I recall explicitly writing it in the former way to avoid point-free style in that section, so that we don't try to cram too many points at once

mikaelbr:
Sure. Makes sense. I was thinking, if you read about functors, you know about partial application, but I get it.

PR #77, opened 2017-08-11
Update mutation.md

philraj:
I'm not sure about this one, but I think this is what was intended. Could be wrong, but either way something is off about the sentence as-is.


chenglou:
Thanks thanks thanks

PR #86, opened 2017-08-17
Upgrade installation instructions for the latest version

anmonteiro:


anmonteiro:
ooops this looks like it hasn't been released yet. I was looking at the wrong place.

PR #89, opened 2017-08-17
Mention reasonml-community/index in guide/javascript/libraries

glennsl:


chenglou:
Geez the index page is such a bleak portrayal of our packages lol

PR #95, opened 2017-08-22
Fix unbound constructor error in example

Tmarty:
Output before:
> Error: Unbound constructor GreetingAndOneSlash

Output after: 
```reason
let greetingAndOneSlash : string = "Hello\nWorld\n\\\nHehe...\n"
```


chenglou:
Oops, thanks!

PR #96, opened 2017-08-22
Fix unbound value error in example

Tmarty:
Rtop output before: 

> Error: Unbound value theChar 

Rtop output after: 
```reson
- : 'a => char => bool = <fun>
```


chenglou:
Thanks!

PR #102, opened 2017-08-23
add companies page with logos of companies using Reason

rickyvetter:
This adds a randomized list of companies to the site that people can add to.


glennsl:
No facebook?

PR #103, opened 2017-08-23
Try Reason fixes

glennsl:
I give no guarantee that any of this actually works as intended, but from my own testing it seems to do what it's supposed to. Buyer beware!

URL length seems to be reduced by about 50% from the current approach.


jaredly:
looks great to me!

PR #108, opened 2017-08-25
Update stdlib (maybe?)

glennsl:
Missing modules have been fixed in bucklescript-playground (see https://github.com/BuckleScript/bucklescript-playground/issues/2)

Did the copy files and run regen script routine, but I couldn't actually get it to work locally. Am I missing a step?


glennsl:
I suppose this is just the js runtime part of it, and I have to rebuild the bs.js bundle to get the external definitions perhaps?

glennsl:
Ok, I think I got it. Just another file that had to be copied and renamed.

regenerateStdlibForBrowser.js
```diff
@@ -21,7 +21,7 @@ try {
   console.log(`
 ** There's an error while reading ${stdlibDir}. Is this your first time setting things up?
 You need to copy the BuckleScript standard library files from here:
-https://github.com/BuckleScript/bucklescript/tree/d21cdfbbe61afb0ae6b79735765de949409947ea/lib/js
+https://github.com/BuckleScript/bucklescript-playground/tree/master/stdlib
```
chenglou: Humm, we're using the commonjs version; this is the amd one though? How does this work?

glennsl: Oh, well it doesn't look like it actually does. I've just been testing with code that didn't need any runtime. If it does need runtime, I get "define is not defined". Hold on, I'll fix.

glennsl:
@chenglou Ok, this seems to work better

chenglou:
Tested and works. Thanks!

PR #111, opened 2017-08-25
[TRANSLATION] Russian guide

slonoed:
Translation for guide section without native part. I improve it when I understand more about native workflow.

It still unclear how to deal with multilang site. So now it just live under `/ru` path.


chenglou:
Small update: @rickyvetter is still working on the translation infra. Sorry for the wait.

rickyvetter:
@slonoed thanks so much for this work! It looks great. Merging in now and will add a language picker and make some small changes shortly.

We want to transition these translations to CrowdIn (https://crwd.in/reason) which will allow us to support translation of inline .js files without any extra maintenance effort moving forward.

slonoed:
@rickyvetter great! Let me know if you need help with it.

PR #112, opened 2017-08-25
Fix typo in let-binding section

danny-andrews:


chenglou:
Aaah oops, this isn't a typo, but you've revealed some bad explanations. I've fixed it in the commit above; thanks!

PR #116, opened 2017-08-27
Update Editor Setup url

scottarnold:
Quick update - looks like the link was pointing to the wrong page.


chenglou:
That's the right page actually! Sometimes we get asked why the editor plugin isn't working; folks forget to install the global bins.

Closing. Thanks for the PR though! =)

PR #124, opened 2017-08-29
fix typo in array get example

viebel:


jaredly:
thanks for the catch!

PR #125, opened 2017-08-29
Various style cleanup/tweaks

jordwalke:
Just some cleanup. Adjusting spacing to make it less cramped. Less contrast as well to make it easier on the eyes. Using the system font for UI (as the Yarn website does).

Before:

screen shot 2017-08-28 at 11 51 22 pm

After:
screen shot 2017-08-28 at 11 51 38 pm


jaredly:
🤷‍♂️ sounds good to me!

PR #126, opened 2017-08-29
Update to latest bucklescript

glennsl:


chenglou:
Tested. Looks good, thanks!

PR #128, opened 2017-08-29
Add playground examples, revamp design and layout a bit

### LINKED ISSUE #120: Try Reason page should have several examples to try out/start from

glennsl:
It might now look a bit like this:
screen shot 2017-08-30 at 1 14 34

Fixes #120


kennetpostigo:
Looking good!

src/pages/try.js
```diff
@@ -21,6 +22,42 @@ if (typeof navigator !== 'undefined') {
     </div>
   )
 }
+
+const examples = [{
+  code: 'C4TwDgpgBMBOHQLxQHIHsAm0CWA7YM80cCUAPlADIQCGAZgNwBQTANhAfAMZQDOArgFsoiJlCh1+uclVp0RAPigAGMePUV0WKADcarftHZ1O2AOYALAogVr16vQegBqKAAoBw48ACUUVx5CULDmVj7MbBxQgiAAKkQiTACQblrQAIx26qmY0ABMWfY52gAssvTldD6FUClpUABslZU+1fbixdAAzDUd9QCszdT0be19uVAA7ENyreEs7ARufqJJntFxCWRKAFK8AHSsaGYMQA',
```
jaredly: let's have the code be in plain text for maintainability, and then convert it w/ js.
e.g.
```
const examples = [{
...
}].map(example => ({name: example.name, code: compress(example.code)}))
```
or something

jaredly: or wait I don't think we need them to be compressed at all?

src/pages/try.js
```diff
@@ -293,7 +379,27 @@ export default class Try extends Component {
...
+              css={styles.toolbarCheckbox}
+              type="checkbox"
+              checked={this.state.autoEvaluate}
+              onChange={this.toggleEvaluate}
+            />
+          </div>
+          <ShareButton
+            url={this.state.shareableUrl}
+            onClick={this.copyShareableUrl}
+          />
```
jaredly: I like the share button!

src/pages/try.js
```diff
@@ -467,23 +561,141 @@ const styles = {
...
     right: 0,
-    backgroundColor: 'rgba(200, 200, 200, 0.6)',
-    padding: '5px 10px',
-    fontWeight: 'bold',
-    fontFamily: 'monospace',
-    color: 'black',
-    fontSize: 14,
-    lineHeight: '14px',
+    backgroundColor: 'rgba(246, 244, 244, 0.8)',
+    padding: '1em',
```
jaredly: I'd prefer `.5em` or so -- there's currently just a ton of whitespace

src/pages/try.js
```diff
@@ -467,23 +561,141 @@ const styles = {
...
     fontFamily: headerFontFamily(),
     fontSize: 16,
-    padding: '10px 20px 0 20px',
+    flexDirection: 'row',
+    justifyContent: 'flex-end'
+  },
+
+  toolbarButton: {
+    borderLeft: '1px solid #d6d4d4',
+    padding: '1em 2em',
```
jaredly: here too, so much whitespace! but 🤷‍♀️

src/pages/try.js
```diff
@@ -467,23 +561,141 @@ const styles = {
...
+    display: 'none',
+    top: '100%',
+    left: 0,
+    minWidth: '100%',
+    width: '25vw',
+    boxShadow: '1px 1px 1px rgba(0, 0, 0, .2)',
+    borderTop: '1px solid #d6d4d4',
+
+    '& li': {
+      padding: '.5em 2em',
```
jaredly: ⬜️ ⬜️

src/pages/try.js
glennsl: I was originally going to just make these ordinary links, so you could move back and forward in history, but it gatsby doesn't seem to pick up on changes in just the query param. But it made sense to have ti compressed then, now not so much. I'll fix.

glennsl:
@jaredly I've now made the examples plaintext

Regarding whitespace, I've played around with it a bit more, but I think it looks best as it is. The vertical padding on menu items is even less than I'd like (and what it originally was), as I now find it a bit difficult to distinguish between separate items and one item spanning multiple lines. But there's so many examples in the menu I had to reduce it.

chenglou:
Let's just make sure it looks good on an 11inch screen

chenglou:
There seem to be a race condition. If you have `evaluate` on, it flashes the result then say `require` is not found

chenglou:
Padding _might_ need adjustment, but let's merge this. This is quality stuff

PR #129, opened 2017-08-30
Modify URL to path

codiemullins:
Using relative path, because with full URL instead of navigating to new page it attempts to navigate to `https://reasonml.github.io/guide/https://reasonml.github.io/reason-react/`


chenglou:
Nice catch! I'm changing it to the codebase convention of appending `//reasonml.github.io` for reason-react. Otherwise locally you see `localhost:8000/reason-react` which 404s.

Thanks!

PR #92, opened 2017-08-18
Add description of .cm* files to FAQ

af:
As a newcomer I found references to these files in compiler/build errors quite cryptic. Not sure if this is the best place to document them but it'd be nice to have this information accessible and searchable.

PR #130, opened 2017-08-30
Integrate facebook/reason/wiki file extension contents into the FAQ

af:
Continuation of #92 

This adds more extensive discussion of file extensions, using content from [the Reason wiki page](https://github.com/facebook/reason/wiki/OCaml-Ecosystem-Extensions-List).


jaredly:
Awesome, thanks for the detailed explanations!

PR #136, opened 2017-08-31
Add comparisons to Elm and PureScript to the FAQ

glennsl:
I don't exactly expect this to go through sans discussion, so have at it! It's also been quite a while since I evaluated them, so there's probably plenty of stuff here that's just plain wrong even.

### DIFF
--- a/src/pages/community/faq.md
+++ b/src/pages/community/faq.md
@@ -23,6 +23,12 @@ They'll take your code from the (macOS) clipboard, convert it, and paste it back
 #### What's the relation between Reason, BuckleScript and OCaml?
 See [here](/guide/javascript). Reason's a syntax for OCaml and supports all its features. BuckleScript compiles OCaml/Reason code into JavaScript.
 
+#### How does Reason (and BuckleScript) compare to Elm
+Reason is a general-purpose language that can target node, native code and utilize both the opam and npm ecossytems. Elm is more opininated and focused, aiming at improving the experience in its own niche and ecosystem, at a rather high cost to its flexibility. Elm has better error messages, though Reason is closing in fast. BuckleScript has superb js interop, generates highly readable and easily debuggable js code and can utilize the npm ecosystem to its full extent, while Elm's FFI and interop is rather convoluted. Elm is pure, while Reason is pragmatic (take that as you wish :)). OCaml has had 25 or so years to mature, and has an active academic base of contributors that keep it close to the forefront of progamming langauge development, while Elm is relatively new and unstable, but moves faster. Reason has first-class support for React, while Elm is focused solely on "The Elm Architecture" (TEA). Reason does however also have a TEA implementation that does a surprisingly good job of mimicing its muse.
+
+#### How does Reason (and BuckleScript) compare to PureScript
+Like Elm, PureScript is relatively new and unstable, more so than Elm. It also targets the JavaScript niche, but unlike Elm it doesn't focus solely on the front-end. Like BuckleScript, PureScript compiles to highly readable JavaScript code, and has significantly better FFI than Elm, though its purity does impose some limitations. PureScript also targets the bower ecosystem instead of npm. Given its name, PureScript is naturally quite heavily invested in its purity, while Reason is and will always be a pragmatic lanaguage. PureScript has several UI frameworks with no clear winner, some of them using or having semantics very close to React, while Reason of course is quite heavily focused on the latter.
+
 #### Where do all these `print_endline`, `string_of_int` functions come from?
 They're from the standard library, pre-`open`ed during the compilation of your file. This is why you see them in scope.
 
--- a/src/pages/community/faq.md
+++ b/src/pages/community/faq.md
@@ -24,10 +24,19 @@ They'll take your code from the (macOS) clipboard, convert it, and paste it back
 See [here](/guide/javascript). Reason's a syntax for OCaml and supports all its features. BuckleScript compiles OCaml/Reason code into JavaScript.
 
 #### How does Reason (and BuckleScript) compare to Elm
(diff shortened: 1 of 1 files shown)


src/pages/community/faq.md
```diff
@@ -23,6 +23,12 @@ They'll take your code from the (macOS) clipboard, convert it, and paste it back
...
 
+#### How does Reason (and BuckleScript) compare to Elm
+Reason is a general-purpose language that can target node, native code and utilize both the opam and npm ecossytems. Elm is more opininated and focused, aiming at improving the experience in its own niche and ecosystem, at a rather high cost to its flexibility. Elm has better error messages, though Reason is closing in fast. BuckleScript has superb js interop, generates highly readable and easily debuggable js code and can utilize the npm ecosystem to its full extent, while Elm's FFI and interop is rather convoluted. Elm is pure, while Reason is pragmatic (take that as you wish :)). OCaml has had 25 or so years to mature, and has an active academic base of contributors that keep it close to the forefront of progamming langauge development, while Elm is relatively new and unstable, but moves faster. Reason has first-class support for React, while Elm is focused solely on "The Elm Architecture" (TEA). Reason does however also have a TEA implementation that does a surprisingly good job of mimicing its muse.
```
OvermindDL1: > Reason does however also have a TEA implementation that does a surprisingly good job of mimicing its muse.

Yay!  ^.^

af: Two suggestions on the Elm section:

* Might as well link to https://github.com/OvermindDL1/bucklescript-tea :)
* "interop is rather convoluted" might be interpreted as combatative– even though I would personally agree with the statement, it might be more diplomatic to use a relative phrase like "is more involved" or "is more complex"

src/pages/community/faq.md
```diff
@@ -23,6 +23,12 @@ They'll take your code from the (macOS) clipboard, convert it, and paste it back
...
+
+#### How does Reason (and BuckleScript) compare to PureScript
+Like Elm, PureScript is relatively new and unstable, more so than Elm. It also targets the JavaScript niche, but unlike Elm it doesn't focus solely on the front-end. Like BuckleScript, PureScript compiles to highly readable JavaScript code, and has significantly better FFI than Elm, though its purity does impose some limitations. PureScript also targets the bower ecosystem instead of npm. Given its name, PureScript is naturally quite heavily invested in its purity, while Reason is and will always be a pragmatic lanaguage. PureScript has several UI frameworks with no clear winner, some of them using or having semantics very close to React, while Reason of course is quite heavily focused on the latter.
```
jaredly: - maybe replace "better FFI than Elm" with "easier/less-strict FFI than Elm"
- "the bower ecosystem (which is [not recommended](https://github.com/bower/bower.github.io/commit/12566e5b777ef4c93c525cf3d2fa35ebc9f4d2e5#diff-c35334318579de7b9e052bd7a04ad054R57) by its maintainers)"
- also the pragmatic thing. Dunno if "Given its name, PureScript is naturally quite heavily invested in its purity" adds a ton. maybe "PureScript is pretty much an attempt to bring Haskell to the web (Elm is inspired by Haskell, but diverges in several key areas), and so will be familiar to people coming from Haskell backgrounds. Reason is literally OCaml, but we're putting in a lot of effort to make it accessible to people with little or no functional programming background."

chenglou:
Let's hold this for a little btw. I'd like to make sure we shine a good light on our cousin languages

src/pages/community/faq.md
glennsl: @jaredly 
* I've tried to integrate your first point with some other points I was making, while being a bit briefer.
* Immutable vs mutable data types is covered by the "pure vs non-pure/pragmatic" point I think, which I realize you'd rather remove, but this is a bit wordy and specific, without addressing the other aspects of purity (which will quickly get you into controversial territory).
* "pragmatic" is admittedly a bit of a weasel word, but Andre completely evacuates the meaning of the words (maybe that was the point though). My intention here is to address what is a very important distinction, acknowledge that the topic is controversial, and at the same time keep it brief. That has to come at some expense to precision. I think being vague is ok as long as we're honest and open about it. It can probably be improved, but I don't think it should just be removed.

src/pages/community/faq.md
glennsl: * I replaced "easier FFI" with "simpler FFI"
* The bower maintainers seem to primarily address js consumers. And the choice of bower is very deliberate from what I understand: http://harry.garrood.me/blog/purescript-why-bower/
* Your last suggestion is a bit wordy, and goes into the motivations of the PS developers which is a bit of a shaky ground I think, familiarity is an important point I've missed. How about just "Reason, more so than OCaml, aims to be familiar to JavaScript developers, while PureScript is much more flirtatious towards Haskell developers". I'd like to make a similar comparison with Elm, but familiarity doesn't really seem to be a concern for them, even if accessibility very much is, which is a bit contradictory IMO.

src/pages/community/faq.md
```diff
@@ -23,6 +23,21 @@ They'll take your code from the (macOS) clipboard, convert it, and paste it back
...
+* Elm is pure, while Reason is "pragmatic" (take that as you wish :)).
+* OCaml has had 25 or so years to mature, and has an active academic base of contributors that keep it close to the forefront of progamming langauge development, while Elm is relatively new and unstable, but moves faster and has less legacy cruft.
+* Reason has first-class support for React, while Elm is focused solely on "The Elm Architecture" (TEA). Reason does however also have [a TEA implementation](https://github.com/OvermindDL1/bucklescript-tea) that does a surprisingly good job of mimicing its muse.
+
+#### How does Reason (and BuckleScript) compare to PureScript
+* Like Elm, PureScript is relatively new and unstable, more so than Elm. It also targets the JavaScript niche, but unlike Elm it doesn't focus solely on the front-end.
+* Like BuckleScript, PureScript compiles to highly readable JavaScript code, and has a significantly simpler FFI system than Elm, though its purity does impose some limitations.
+* PureScript also targets the bower ecosystem instead of npm.
```
bassjacob: this isn't necessarily true anymore, but also isn't as much an issue for purescript. it isn't use bower to replace npm, it just doesn't allow nested resolution of modules so I'm not sure that it matter for the conversation?

glennsl: Huh, TIL, do you have a source that expands on that?

PR #138, opened 2017-09-01
Fix alignment/spacing of logo

glennsl:
It used to look like

screen shot 2017-09-01 at 4 53 41

and now it looks like

screen shot 2017-09-01 at 4 53 22


chenglou:
Hah, block =P
Good initiative to get rid of unfortunate flex box drawbacks

PR #140, opened 2017-09-01
Move playground evaluation into a web worker, terminate on timeout

### LINKED ISSUE #131: Playground eval timeout to avoid the infiniteness of infinite loops
Auto-evaluating an accidental infinite loop caused by some intermediate editing state kind of sucks. Not just because the whole tab freezes, but you'll also lose everything you did up to that point. Even with the new localStorage, a kill & refresh will just land you straight back into the same infinite loop of suckiness.

So yea.. Timeout good.

glennsl:
Fixes #131 

This will need some more testing and tweaking of thresholds I think. The timeout is currently set to just 1 second, and I've also added a logging limit currently set to 100 items, but there's still some freezing on my end. Question is whether it's acceptable.

### DIFF
--- a/src/pages/try.js
+++ b/src/pages/try.js
@@ -75,7 +75,7 @@ const errorTimeout = 500
 
 const waitUntilScriptsLoaded = done => {
   const tout = setInterval(() => {
-    if (window.refmt && window.ocaml && window.require) {
+    if (window.refmt && window.ocaml) {
       clearInterval(tout)
       done()
     }
@@ -129,6 +129,14 @@ export default class Try extends Component {
         this.output(data);
       }
     }
+    this.evalWorker.onerror = err => {
+      this.errorTimerId = setTimeout(
+        () => this.setState(_ => ({
+          jsError: err
+        })),
+        errorTimeout
+      );
+    }
   }
 
   evalJs(code) {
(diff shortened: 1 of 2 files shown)


chenglou:
Looks good! Ready to merge?

glennsl:
If you've tested it and feel confident about it, sure!

chenglou:
Just checked, this gives the following error:

screenshot 2017-09-06 02 22 48

Any idea what's happening here? Repro:

```reason
let a=List.map (fun a=>a)[1,2,3];
```

glennsl:
Ah, yes it's because of some stale artifacts that were included in the bs compiler build. It has nothing to do with this PR and I don't think it really is a problem in practice, but bucklescript-playground has an updated build where this is fixed.

chenglou:
You sure? I don't see it in current tip

glennsl:
It's probably just because putting eval in a web worker doesn't suppress warnings somehow :) I explicitly fixed this issue in bucklescript-playground, just forgot to push the new one here too.

glennsl:
Wait, that doesn't make sense... it's bs compilation that warns, not eval...

glennsl:
Ah, it's because the console.error is no longer overridden since output is now captured in the web worker instead.

chenglou:
Aaaah kk. cc @anmonteiro 

This looks good, thanks!

PR #143, opened 2017-09-02
sidebar readability tweaks

glennsl:
Not 100% sure about the gray background, but otherwise I find the sidebar much more readable now.

Before:
screen shot 2017-09-02 at 3 52 59

After:
screen shot 2017-09-02 at 3 52 42

### DIFF
--- a/src/components/GuideSidebar.js
+++ b/src/components/GuideSidebar.js
@@ -2,16 +2,16 @@ import React from 'react'
 import Link from './Link'
 
 import {scale, rhythm, headerFontFamily} from '../utils/typography'
-import {text, accent, dividerLine} from '../utils/colors'
+import {text, accent, dividerLine, gray} from '../utils/colors'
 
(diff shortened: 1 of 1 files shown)


chenglou:
I see that on longer pages it looks like this:

screenshot 2017-09-05 17 11 08

Is this intentional?

glennsl:
It has minHeight 100%, but maybe safari issue?

chenglou:
Yeah safari issue...

chenglou:
I think that's fine enough. Also, I was thinking maybe the side bar should be scrollable and fill page height anyway. Maybe later

glennsl:
Definitely, I'm just lazy ;) I'll look into the safari issue too.

chenglou:
I got paralyzed by the challenge. Most sidebar implementation either doesn't have bounce back (I NEED BOUNCEBACK) or has scrolling glitches in one section you scroll the other

PR #145, opened 2017-09-02
Embedded playground (proof-of-concept)

glennsl:
screen shot 2017-09-02 at 9 23 24

This is a very quick and dirty proof-of-concept without any lazy loading, error handling, sandboxing or anything else beyond the basic needed to edit and run code. Before I venture any further I wanted to get some review and feedback on the technical approach and design. I'd also like to wait until the other playground PRs are merged so I can refactor the common parts out.

### DIFF
--- a/src/pages/guide/language/overview.md
+++ b/src/pages/guide/language/overview.md
@@ -3,6 +3,17 @@ title: Overview
 order: 0
 ---
 
+```playground:reason
+let rec hanoi n a b c =>
+  if (n > 0) {
+    hanoi (n - 1) a c b;
+    Js.log {j|Move disk from pole $a to pole $b|j};
+    hanoi (n - 1) c b a
+  };
+let () =
+  hanoi 4 1 2 3;
+```
+
 Primitive                             | Example
 --------------------------------------|--------------------------------
 Strings                               |  `"Hello"`
--- a/src/pages/guide/language/overview.md
+++ b/src/pages/guide/language/overview.md
@@ -10,6 +10,7 @@ let rec hanoi n a b c =>
     Js.log {j|Move disk from pole $a to pole $b|j};
     hanoi (n - 1) c b a
   };
+  
 let () =
   hanoi 4 1 2 3;
 ```
--- a/embedded-playground/index.js
+++ b/embedded-playground/index.js
@@ -1,0 +1,22 @@
+const visit = require(`unist-util-visit`)
+
+module.exports = ({ markdownAST }) => {
+  visit(markdownAST, `code`, node => {
+    let {lang} = node;
+
+    if (!lang || !lang.startsWith('playground:')) {
+      return;
+    }
+
+    node.type = `html`;
+    node.value = 
+      `<div class="embedded-playground">
+        <div class="editor-container"><pre>${node.value}</pre></div>
+        <div class="output"></div>
+        <div class="buttons">
+          <button class="run-button">Run</button>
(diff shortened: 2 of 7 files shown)


glennsl:
Added a reset button and tweaked the appearance a bit:
screen shot 2017-09-02 at 11 37 15

src/pages/guide/language/overview.md
```diff
@@ -3,6 +3,18 @@ title: Overview
...
+    hanoi (n - 1) c b a
+  };
+  
+let () =
+  hanoi 4 1 2 3;
+```
+
 Primitive                             | Example
 --------------------------------------|--------------------------------
 Strings                               |  `"Hello"`
```
chenglou: This page needs more update. It's a rather weak table for an overview

glennsl: I'll fill it in e a bit more before this is merged.

glennsl:
> Good to merge?

Not even close :) It needs some heavy refactoring and attention to non-happy paths. And I need the other playground PRs merged first.

src/pages/guide/language/overview.md
glennsl: Also, feel free to add to this PR if there's something you'd like to have here.

chenglou: Ah don't worry about this one, I'll change it myself at one point

viebel:
@glennsl would you like to join forces?
Are you familiar with my Klipse plugin https://github.com/viebel/klipse?
Now, it supports reason http://blog.klipse.tech/reason/2017/10/06/blog-reason.html

glennsl:
@viebel Hi, that's awesome! This prototype already does that, however. You can even selectively choose which snippets to make interactive by marking them `playground:reason` in the markdown (as opposed to just `reason`). And it does so while reusing most of the existing infrastructure for the playground (the reason why it's not been merged yet is mostly because I want to refactor the playground to be able to reuse even more). Are there any other benefits to using klipse instead?

A few impressions from trying it out: It looks like it uses the same approach as the playground with refmting Reason to OCaml code before compiling the latter, and therefore also suffer from bugs in the OCaml pretty printer. It also doesn't seem to handle infinite loops and overwhelming output very elegantly (either just not doing anything, or locking up until it's done, respectively), and it doesn't seem to have a "reset" button, though I'm sure that wouldn't be too hard to add.

chenglou:
There's probably value using this in third-party blog posts (imagine dynamic refmt-ing to the latest version of reason syntax) too, so dog fooding klipse might be a good idea

kennetpostigo:
I'm definitely using Klipse in the future, didn't know it existed.

viebel:
@glennsl another issue that we have in Klipse is that code and data cannot be shared between snippets as each snippet is compiled separately (do you know how to fix it?).

The big difference between Klipse and the prototype exposed in this PR is that Klipse is a 3rd party solution that can be embedded into any web page, blog post, documenation etc...

viebel:
@chenglou what do you mean by "dog-fooding" Klipse?

chenglou:
@viebel I mean using it in at least one of our doc sites. The situation is that right now, if we do any syntax upgrade, we have to go and update all the docs and prominent blog posts, lots of which are from external contributors. If we have folks use a no-fuzz, lean, dynamic code block instead of static code, then these would be automagically upgraded once they e.g. update klipse and get the new reason syntax or something.

PR #146, opened 2017-09-03
Update playground links to use https://reasonml.github.io/try/

jchavarri:
Thanks for these docs! And Reason. Amazing and very exciting stuff. ✨ 

Just noted while going through the docs that a couple of links to the BuckleScript playground were dead (`https://bucklescript.github.io/bucklescript/js-demo` seems to have been moved to `https://bucklescript.github.io/bucklescript-playground`). I thought to just update the urls but then... why not use the brand new Reason playground? :)


chenglou:
Welcome! And thanks. I forgot to change these after we revamped our playground.

We use `/try/foo` in the docs, no need to specify the `reasonml.github.io` prefix (which might change in the future), unless it's a `reason-react` link. I've patched that on top of your contribution.

Thanks again!

PR #149, opened 2017-09-05
Update global-installation.md

hew:
Updates instructions to account for symlinks that occur during npm/yarn installs.


chenglou:
Ah yeah good point. Thanks =)

PR #151, opened 2017-09-06
Fix command for watch mode in quickstart

neeharv:


chenglou:
Hah, thanks! Though this one was supposed to be `start`. I'll fix it in the bsb template.

PR #154, opened 2017-09-06
Added examples to the list API docs

### LINKED ISSUE #152: Code examples in API Docs
Is this something you'd be interested in having me add, or is the point of the API docs section to be concise? I was thinking maybe a simple, 3-4 line max example of each function listed.

JakeDawkins:
**Resolves #152.** 

I mentioned this in #152, but I had some extra time today, so I documented the list API.

All I had to do was...
1. add a class for code examples in the css
2. add a `code` block with the samples in it. 
3. I also changed the list syntax in the docs from `[a; b; c]` to `[a, b, c]` to follow reason syntax instead of OCaml.

**Screenshot:**

screen shot 2017-09-06 at 2 34 18 pm

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "editor.formatOnSave": false
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-  "editor.formatOnSave": false
-}
--- a/src/pages/api/List.html
+++ b/src/pages/api/List.html
@@ -188,61 +188,94 @@ List.fold_right (fun v a => a + v ) [0, 1, 2] 0;  /* returns 3 */
 </pre>
 </div>
 <br>
+
 <h6 id="6_Iteratorsontwolists">Iterators on two lists</h6><br>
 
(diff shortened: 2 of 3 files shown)


chenglou:
@jsdf how easy would it be to maintain this?

JakeDawkins:
If this would be unmaintainable, I can just publish the examples somewhere. I realize having this on only _some_ of the api docs would be unfortunate. I'm also not sure how you handle changes in this project in respect to the ocaml docs.

jsdf:
@JakeDawkins sorry for the delay in responding. These HTML files are generated from the OCaml standard library by the scripts here: https://github.com/reasonml/reasonml.github.io/tree/source/generate-docs

So editing these HTML files directly means we wouldn't be able to regenerate the docs HTML from OCaml's files without losing your contributions. Ideally you would contribute these changes (but with OCaml syntax) to the relevant files in the upstream OCaml project here: https://github.com/ocaml/ocaml/blob/trunk/stdlib/list.mli#L51-L57

Alternatively we could try to make a change to the [script which converts the docs to Reason](https://github.com/reasonml/reasonml.github.io/blob/source/generate-docs/transform.js) to add in extra Reason-specific stuff after each item.

PR #156, opened 2017-09-09
Include Fable(F#) in the list of alternatives

basbz:
I consider F# and OCaml to be closely related and think it strange to leave it out.
It's also more or less based on the javascriptair episode on "Typed Functional Programming in JavaScript" : https://javascriptair.com/episodes/2016-08-03/


chenglou:
Thanks!

PR #161, opened 2017-09-11
Added Rust as close cousin ;) Also Ocaml based, LLVM driven. Also great language :)

dmilith:


OvermindDL1:
Not sure rust is close though?  Rust has *much* more in common with Haskell style languages than it does anything even remotely near ML style languages (like Reason and OCaml).  Though I really really wish Rust did...

glennsl:
Rust and Reason isn't really targeting the same audience either. Reason is currently focusing on compile-to-js, and the other alternatives reflect that (hence Fable being an alternative instead of F#, and PureScript instead of Haskell). As far as I know there isn't a good compile-to-js story for Rust yet (there's been a few attempts to get it to work with emscripten, but nothing substantial I think).

jaredly:
🤷‍♀️ rust is cool, it's another language that "developers interested in cool ~new languages" might be into.

glennsl:
I love Rust! I think it's exceptionally well-designed for its intended use. That still doesn't make it an alternative to Reason. If it was I probably wouldn't even be here.

PR #165, opened 2017-09-14
Add an example of using Js.Nullable.to_opt

robbiemccorkell:
It took me quite a while to find a nice way of handling possibly null JS values. 

Highlighting the use of Js.Nullable.to_opt might save people some time.


chenglou:
Good point. I've reworded things a little.

Thanks!

robbiemccorkell:
No problem, thanks for tweaking it.

PR #166, opened 2017-09-14
Update reason-cli version

Erwyn:


Erwyn:
I was not sure if it's correlated or not, I suppose it is if you say this.

chenglou:
Thanks!

PR #168, opened 2017-09-15
Update JavaScript/Reason tuple comparison

yawaramin:
Showcase Reason tuples by using multiple types, and show how to simulate
tuples in JavaScript using arrays.


chenglou:
Fair comparison =)

PR #173, opened 2017-09-20
Remove explicit watch mode command

maartenvanvliet:
The explicit `npm run watch` command was removed in this commit: https://github.com/BuckleScript/bucklescript/commit/c8778eddc57467998715f91ea00edaee1ad141ec#diff-7b08ae4b16cddfb4e8606ad3e909ef23
Instead the `npm run start` command runs in watch mode


chenglou:
Thanks!

PR #177, opened 2017-09-24
Make header mobile button visible

CharlesMangwa:
The new header behaviour is great on mobile! The only issue is that its orange background makes it invisible on all pages, except the homepage. This PR fixes that.

### DIFF
--- a/src/components/HeaderNav.js
+++ b/src/components/HeaderNav.js
@@ -1,8 +1,8 @@
 import React from "react"
 // import LanguageSelect from './LanguageSelect'
 import Link from "./Link"
-import {accent} from "../utils/colors"
-import {headerFontFamily} from '../utils/typography'
+import { accent, gray } from "../utils/colors"
+import { headerFontFamily } from '../utils/typography'
 
 const links = [
   {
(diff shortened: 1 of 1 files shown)


patcito:
That's weird as it was working on my branch probably cause it was based on master instead of source, looks like Cheng fixed it though here :) https://github.com/reasonml/reasonml.github.io/commit/bc0bfff43f93cc98a801ddf027b340bbb688d13f

chenglou:
Sorry, I'm not sure I understand this; what orange background?

Also, after tweaking the mobile style, I realized that we might be better off keeping the horizontal nav bar, with smaller front & horizontal scrolling

CharlesMangwa:
Here is how it looks like before the PR:
orange header with orange mobile navbar

and after:
orange header with white mobile navbar

because the 3 little bars always have that "orange background". However: I do prefer you option with an horizontal nav bar, seems much more easier to use to me.

chenglou:
I've tried the scrollbar and didn't like it. Two rows is plenty clear:

localhost-8000- iphone 6

chenglou:
@CharlesMangwa I've made our temporary deploy easier. The new `new-deploy` branch just simply removes the Russian and Spanish folder for now. You shouldn't need any other hack to make things work in the meantime.

CharlesMangwa:
@chenglou Alright got it, I'll check your modifications. I've managed to put deploy

CharlesMangwa:
We chose another system (29852097bc3c2d1fc9006191195e9bbc42b2c876) so closing this PR.

PR #179, opened 2017-09-25
Resuggest integration

GuillaumeSalles:
@chenglou [proposed](https://github.com/GuillaumeSalles/resuggest/issues/6) to show [resuggest](https://guillaumesalles.github.io/resuggest/) on the reasonml doc. 

So as a starting point, I just copy pasted the whole code source (except the unit tests) and did a small refac to reuse some of the styles / utils. It lacks documentation and a link to https://reasonml.github.io/resuggest/. Also, I'm not sure how Gatsby bundle the js files, but it's probably not a good idea to add 'bs-platform' as dependency.

Maybe it will be easier to extract the suggest function in it's own repo/package?

Waiting for your feedback before going further! Thanks

### Screenshot

screen shot 2017-09-24 at 5 30 14 pm


chenglou:
Awesome work! =D

We might delay this integration slightly, or do it soon, depending on how @rickyvetter's doc generator exploration goes.

GuillaumeSalles:
No problem! Let me know if I can do something

GuillaumeSalles:
I updated resuggest a little bit to support Reason 3+ and now it supports most of the pure functions of the standard library. I also added a link to try a suggestion in the playground. 

Still interested to move it to the official doc? Or do you prefer to let it leave outside? I don't really have a preference, I just don't like to let my PRs open for 1 year 🙂

Side note, do you think it's the right time to add Belt to resuggest or is it too soon?

chenglou:
We should definitely add Belt, yes. Sorry about the open PR. I'm wondering if this can be integrated (without too much burden) into @jaredly's https://github.com/jaredly/docre. This way you can have it per library, which is massively useful

jaredly:
this is cool for finding stdlib functions for string manipulation, but I question how useful it would be for libraries in general 😅.
For example, if you're trying to find "ReasonReact.string", and you put `"hello"` in the input, and... manually construct a react element in the output?

chenglou:
@jaredly yeah for things that return an abstract type, it's a bit hard. Though `ReasonReact.string` is a special case. Basically this amounts to whether we can find an adequate default comparison function, which I think we actually can lots of times. Kinda useful for side-effectful APIs too sometimes? (searching for `()`)

chenglou:
Actually, regarding the above: I think that a string-based, normalized (aka refmt-ed) comparison would work? So if you search for the return value `[1,2,3]` it'll refmt to `[1, 2, 3]` then you compare strings

PR #188, opened 2017-10-03
Debounce search

### LINKED ISSUE #52: Debounce search keystrokes

JoshYork:
Resolves #52. I set it to debounce by 200ms which seemed about right when testing.


chenglou:
Thanks!

PR #190, opened 2017-10-03
revamp community page

glennsl:
Taking a page (quite literally) from the new reactjs site. Not everything about this is terribly serious, but with some input from y'all I'm sure we can make it glorious!


chenglou:
Hah that's intense lol. I'll review it later

PR #191, opened 2017-10-04
Add a sub-section on extending module signatures

alexeygolev:
Following the example for extending modules.
I assume with new syntax update it will be `include (module type of ...). Maybe this PR can wait for the new syntax and I can update it when it's been released.


chenglou:
No need to wait. We'll update the whole doc to the new syntax later anyway. Thanks!

PR #193, opened 2017-10-07
Fix the example shown on the front page

fazouane-marouane:
The example on the [front page](https://reasonml.github.io/) always bugged me. It turns out it contains an error (or is incomplete): `Unbound value stranger`.
I suppose it was meant to be a function definition.


chenglou:
Thanks!

PR #201, opened 2017-10-11
Update global-installation.md

mdedetrich:
There is another reason why Visual Studio Code can report an Unbound Module, its if you haven't built your project with npm yet


chenglou:
Thanks! Tweaked the message

PR #206, opened 2017-10-16
Fix ReasonReact link

nolds9:
Current absolute link is not resolved. Uses relative path for ReasonReact docs.


chenglou:
Thanks! I'll hold off this one. People have been changing this link back and forth for a while now and it seems very confusing which one is handled. cc @rickyvetter

chenglou:
K, prefixed with `//` instead, because apparently that's the idiom. Thanks!

PR #212, opened 2017-10-26
Fix Exception.md example type error

bobbypriambodo:
The current example does not type check ([try](https://reasonml.github.io/try/?reason=DYUwLgBA5uCSYgLYQBRgBYgDIEsDOkAXBMPpDgHZgCUEAvAHwQDeAUBBDgGaq4EB0oClAwQM2MhCYAGWmw4c+YfugAmYzEvYQAvhBDA8IFto4AnAIb5jAOQD2YAPpc7AVwqrtO1joDcrVlBIMxA8V2BIOjEzAE9UGDB4JAgAbQBGABoIACYsgGYAXTlWAB8IeycXd3VGCAAHM0onEA9SCmMAIiTkCgcIKo8AQg6fXyA)). `result` expects `int` (the item) but is given `unit` from `print_endline`.

The proposed change (with `failwith`) type checks ([try](https://reasonml.github.io/try/?reason=DYUwLgBA5uCSYgLYQBRgBYgDIEsDOkAXBMPpDgHZgCUEAvAHwQDeAUBBDgGaq4EB0oClAwQM2MhCYAGWmw4c+YfugAmYzEvYQAvhBDA8IFto4AnAIb5jAOQD2YAPpc7AVwqrtO1joDcrVlBIMxA8V2BIOjEzAE9UGDB4JAgAbQBGABoIACYsgGYAXTlWAB8IeycXd3VGCC4rYAB3HFEAIiTkCgc6tw8AQlafXyA)), but not sure if rethrowing is acceptable as an example.

We might want to show how "catching" exceptions can return a value, I think it's a great feature to show off ([link to tweet](https://twitter.com/aureliengeorget/status/923150639047348224)). But then again it might encourage people to use exceptions, so I'm not sure on this one too. Would love to update this PR if that's what preferred.

### DIFF
--- a/src/pages/guide/language/exception.md
+++ b/src/pages/guide/language/exception.md
@@ -17,7 +17,7 @@ let getItem theList => {
 };
 
 let result = try (getItem [1, 2, 3]) {
-| Not_found => print_endline "Item not found!"
+| Not_found => failwith "Item not found!"
 };
 ```
 
--- a/src/pages/guide/language/exception.md
+++ b/src/pages/guide/language/exception.md
@@ -17,9 +17,11 @@ let getItem theList => {
 };
 
 let result = try (getItem [1, 2, 3]) {
-| Not_found => failwith "Item not found!"
+| Not_found => 0 /* Default value if getItem throws */
 };
 ```
+
+Note that the above is just for demonstration purposes; in reality, you'd return an `option int` directly from `getItem` and avoid the `try` altogether.
 
 You can directly match on exceptions _while_ getting another return value from a function:
 


chenglou:
Ahh thanks for the catch. How about this?

```
let result = try (Some (getItem [1, 2, 3])) {
| Not_found => None
};
```

But that kinda defeats the purpose and `getItem` should just have returned `option int` instead...

bobbypriambodo:
That makes sense if `getItem` is something that we cannot change (e.g. 3rd party library). Perhaps we can make it seems so? E.g.

```reason
/* ThirdParty.getItem might throw Not_found exception,
   we wrap them to make result an option int */
let result = try (Some (ThirdParty.getItem [1, 2, 3])) {
| Not_found => None
};
```

But that does seem too verbose. I wonder if we can make it more concise.

bobbypriambodo:
Or, perhaps use it for returning default value:

```reason
let result = try (getItem [1, 2, 3]) {
| Not_found => 0 /* Default value if getItem throws */
};
```

chenglou:
Thought of that, except we should encourage people to use an `option` for that instead of doing the JS way and return e.g. `-1` or `0`

chenglou:
Bah, I've updated it to your last snippet and put a note on it afterward. Thanks!

bobbypriambodo:
Neat!

PR #213, opened 2017-10-27
Fixed typo in the "Punning" section

netgusto:


chenglou:
Hello! Actually, the original is right. We do _desugar_ it to `checked={checked}`.

Closing. Thanks for the PR though!

PR #215, opened 2017-10-27
Convert left-over labeled argument docs to the new syntax

jayphelps:
A few places were accidentally left using the previous syntax, and there was a section that is no longer applicable.


src/pages/guide/language/function.md
```diff
@@ -91,16 +93,6 @@ let drawCircle = (~radius as r, ~color as c) => {
...
-For the common case of `radius::radius` (where the label is the same as the local variable name), we have the syntax shorthand `::x`:
-
-```reason
-let drawCircle = (~radius, ~color) => {
-  setColor(color);
-  startAt(radius, radius);
-  ...
-};
-```
-
```
jayphelps: I believe this section is no longer applicable.

chenglou: This is still applicable but the wording needs to change a bit. `(~radius)` is shorthand for `(~radius as radius)`

jayphelps: Okiedoke, I can take a crack at it once I grasp the distinction. Can you clarify the difference conceptually between these:

Reason: `(~radius) =>` is shorthand for `(~radius as radius) =>`
JS: `({ radius }) =>` is not really shorthand for `({ radius: radius }) =>` because it would be conceptually extraneous/redundant

Ignore the fact that passing an object and naming an arg are not the same thing, just referring to the concept of aliasing. As an outsider, the distinction between the two feels extraneous, like in that JS example?

chenglou: Both are just labeled punning for their more verbose counterpart, if that's what you mean?

jayphelps: Gotcha. Mostly I feel this is an unnecessary distinction, and one that actually confuses more than helps--like it would confuse if I was teaching JS and said `({ radius }) =>` was short for `({ radius: radius }) =>`. Obv your call, just feedback.

jayphelps: I'll try to craft an explanation later today, or you're welcome to patch on top. 👍

chenglou: K, removed, thanks!

chenglou:
I've tweaked the wording a bit and merged your diff on top of the other corrections of function arguments people brought. Thanks!

PR #216, opened 2017-10-28
Amend a few leftovers from old syntax.

gandulio:


gandulio:
Ha looks like someone else beat me to it, never mind!

Issue #195, opened 2017-10-08
Instructions on Getting Started with Windows


chenglou:
This should turn into a doc page in the future, but I'd like to gather some feedback on Reason on Windows. In the meantime we'll link this from the doc site.

Context:

Reason on Windows is complicated.

- BuckleScript (bs-platform) 1.9.3 is verified to work well on Windows (1.9.2 had some bugs)
- Reason-cli, through which the important binaries merlin, ocamlmerlin_reason and refmt are installed, doesn't work on Windows.
- OPAM, not verified. But we'll focus on the JS story in this issue.

Workarounds:

- For merlin, a temporary substitute is to have [bsb errors piped into vscode](https://github.com/reasonml-editor/vscode-reasonml#bsb). Bsb is the build system from bs-platform, which works with Windows. Although apparently

superherointj:
**## [WIP. This is a draft I am and will be editing. I'm testing instructions on a clean Windows environment. NOT READY. Please, just wait.]**

yawaramin:
Here is what I did to get OCaml + Merlin working. Haven't tried Reason.

- Install WSL with https://msdn.microsoft.com/en-us/commandline/wsl/install_guide
- Restart
- Install node using the Windows installer
- Add the following to my user account's path:
  - `C:\Program Files\nodejs`
  - `%APPDATA\npm`
- At the Windows command prompt: `npm i -g npm ocaml-on-windows bs-platform`
  - `npm` to get version 5.4.2
  - `ocaml-on-windows` took a while to install
- At the WSL `bash` prompt: `opam switch 4.02.3; eval $(opam config env); opam install merlin`
- Install VSCode & @hackwaly 's OCaml extension
- Add the following to VSCode settings: `"ocaml.merlinPath": "ocamlmerlin.cmd"`
- At the Windows command prompt: `bsb -init test-proj`
- Open VSCode to `test-proj` directory and check out the type hints etc.

This is I think the bare minimum to get OCaml and Merlin working on Windows, thanks to WSL and @fhelwanger 's scripts (the `.cmd` files) mapping each OCaml toolchain command to its corresponding command in the WSL installation.

fhelwanger:
@superherointj Thanks for writing this!

[...]

PR #219, opened 2017-10-28
Update opam installation versions

jdlehman:
The opam installation versions are locked to older versions of merlin and reason. Is there a reason these are locked at these versions or have they just been missed in doc updates?

Removing the version constraints will install the latest, which should be the desired case for most users and will have the added benefit of not needing to update this file for version changes

### DIFF
--- a/src/pages/guide/editor-tools/global-installation.md
+++ b/src/pages/guide/editor-tools/global-installation.md
@@ -22,8 +22,8 @@ order: 10
 ```
 opam update
 opam switch 4.02.3 # mandatory!
-opam install reason.1.13.7
-opam install merlin.2.5.4
+opam install reason
+opam install merlin
 ```
 
 ### Troubleshooting


zploskey:
The instructions should at the very least be updated to reason 3.0.0. I'm not sure if there is any particular reason that merlin 3 won't work, but I noticed that `reason-cli` still ships 2.5.4.

jdlehman:
Agreed. `3.0.0` isn't in `opam` yet last I checked, but I am happy to make the change if we want to be specific (and update the merlin version if there is a reason v3 doesn't work).

zploskey:
I think it makes sense to change it now, even if it before 3.0 is approved in OPAM, because none of the syntax examples will work unless Reason 3 is installed. Depending how long we expect that to take, it might be worth add a note that it isn't there yet. Hopefully it is approved quickly.

jordwalke:
Version 3.0.0 is in opam now, so that's a good change. I think we should hold off on bumping the merlin version until we know that the most recent version of merlin can work well with bucklescript. If someone can confirm, we can also bump the version of merlin.

yawaramin:
Also I'd suggest to provide all OPAM package names in a single install command to help it do better resolution and minimise recompilation. I think this is a best practice in OPAM, see e.g. https://discuss.ocaml.org/t/why-does-opam-recompile-so-much/483/6?u=yawaramin

So, `opam install reason merlin`

jdlehman:
I updated the versions to be hard coded, but I kept the merlin version at `2.5.4` for now. So this should be good to merge.

Based on the [merlin changelog](https://github.com/ocaml/merlin/blob/master/CHANGELOG#L5), it looks like the most notable change in `3.0.3` is Windows support, which might mean that Windows users no longer need to use Windows Subsystem for Linux (WSL). merlin `3.0.3` works locally for me on OSX with reason/bucklescript, but I do not have a Windows machine so I was not able to confirm the Windows support.

If we can get confirmation around Windows support we might be able to make a lot of improvements to docs for Windows users (especially improving the current [installation process](https://github.com/reasonml/reasonml.github.io/issues/195)) in subsequent PRs 😄

chenglou:
We'll verify merlin soon, thanks!

superherointj:
Merlin 3.0.3 is not working on Windows.

PR #221, opened 2017-10-28
fix some pre-v3 function syntax

ccmtaylor:
please note that I've not written any reasonml yet (or OCaml for that matter), so please double-check if I got it right.

### DIFF
--- a/src/pages/guide/language/function.md
+++ b/src/pages/guide/language/function.md
@@ -79,7 +79,7 @@ Since we have currying (more on that below), we can provide the arguments in any
 addCoordinates(~y=6, ~x=5);
 ```
 
-The `x::x` part during declaration means the function accepts an argument labeled `x` and can refer to it in the function body as the variable `x`. This is so that we can have the following pattern, where labeled arguments are renamed inside the function for conciseness:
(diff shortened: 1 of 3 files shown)


src/pages/guide/language/function.md
```diff
@@ -204,7 +204,7 @@ let twelve = addFive(7);
 Actually, the above `add` is nothing but syntactic sugar for this:
 
 ```reason
-let add = (x, y) => x + y;
+let add = (x) => (y) => x + y;
```
zploskey: This was actually correct as it was. Your change with `refmt` to what was already there on reason 3.

ccmtaylor: hmm, I thought this paragraph was explaining that the parameter-list syntax is equivalent to the curried syntax. 

> Reason functions can automatically be **partially** called:
 
>  ```reason
>  let add = (x, y) => x + y;
>  let addFive = add(5);
>  let eleven = addFive(6);
>  let twelve = addFive(7);
>  ```
 
>   Actually, the above `add` is nothing but syntactic sugar for this:
>   
>  ```reason
> let add = (x, y) => x + y;
> ```

note that the definitions of `add` are identical, so the paragraph doesn't make much sense as-is.

zploskey: Aha, you're right, I didn't look at the wider context. That said, this all looks good to me.

chenglou:
Hello! #222 is a superset of this. Symbolically merging this =)

PR #229, opened 2017-11-01
Add ReasonReact support to the playground

astrada:
* Update BuckleScript compiler (bs.js) to 2.0.0
* Add PPX rewriter jsxV2.js (jsoo_reactjs_jsx_ppx_v2.js from Bucklescript/2.0)
* Update setupSomeArtifacts.js to build all the new artifacts
* Add new dependencies:
    - reason-react: to generate bundles for loading ReasonReact modules into the playground
    - bs-platform: to compile reason-react
    - react-async-script-loader: to dynamically load new assets
* Add Preview component (with cleanslate to reset css)


astrada:
This patch is a bit more invasive that I anticipated. Feedback is very welcome. I've tested the new playground on Chrome and Firefox on an Ubuntu 16.04 box (sorry but I don't have Safari/macOS).

astrada:
@jaredly: what do you think about this PR? Have you time to review it? Thanks!

setupSomeArtifacts.js
```diff
@@ -53,3 +54,172 @@ b
...
+// === third task
+
+// Bundle BuckleScript compiler with ReasonReact .cmi and .cmj files:
+// 1. compile ReasonReact with bsb
+// 2. serialize ReasonReact .cmi and .cmj files
+// 3. bundle them with BuckleScript compiler (bs.js)
+
+/** Serialize a .cmi or a .cmj to a string that can be passed to
+  * ocaml.load_module */
+function serializeBinary(binFileName, jsFileName) {
```
chenglou: Damn, this is complex. Where does it come from? If this is mandatory, can you put an elaborate comment?

astrada: Sorry, I forgot to document it. It's based on https://github.com/ocsigen/js_of_ocaml/blob/master/compiler/lib/js_output.ml#L279. I translated it to Reason, compiled with BuckleScript and then manually removed the bs-platform dependencies. I've updated the comment.

astrada:
> I'm eager to see a future playground where we can dynamically load a reason/bs lib!

Testing this patch, I discovered that unfortunately you have to load all the `.cmi`s before using the compiler and that's why I had to bundle the compiler and the `.cmi`s/`cmj`s in `bsReasonReact.js`. I think that `caml_fs_register` doesn't work after initialization. Maybe I can check jsoo source to see if there is some way to enable dynamic loading.

setupSomeArtifacts.js
```diff
@@ -53,3 +54,179 @@ b
...
+  },
+  {
+    cmi: path.join(reasonReactBsDir, 'reasonReact.cmi'),
+    cmj: path.join(reasonReactBsDir, 'reasonReact.cmj'),
+  },
+  {
+    cmi: path.join(reasonReactBsDir, 'reasonReactOptimizedCreateClass.cmi'),
+    cmj: path.join(reasonReactBsDir, 'reasonReactOptimizedCreateClass.cmj'),
+  },
+];
```
jaredly: this could be rather shorter / more DRY as 
```
const moduleNames = ['reactDOMRe', 'reactDOMServerRe', 'reactEventRe', 'reasonReact', 'reasonReactOptimizedCreateClass'];
const modules = moduleNames.map(name => ({cmi: name + '.cmi', cmj: name + '.cmj'}))
```
don't feel too strongly though

src/pages/try.js
```diff
@@ -384,9 +426,11 @@ export default class Try extends Component {
 
     this.setState((prevState, _) => {
       let newOcamlCode = prevState.ocaml;
+      let newShowPreviewPanel = prevState.showPreviewPanel;
       try {
         newOcamlCode = window.printML(window.parseRE(newReasonCode))
-        this.tryCompiling(newReasonCode, newOcamlCode)
+        newShowPreviewPanel = this.detectReasonReact(newOcamlCode);
```
jaredly: good call with the auto-show here

setupSomeArtifacts.js
astrada: I like it!

src/pages/try.js
astrada: thanks!

chenglou:
Tried to bring this in but there are some errors. JSX rewriter not available globally and ReasonReact still can't be found. I'll check later.

astrada:
This patch needs to be updated after upgrading bucklescript compiler. I will try to fix it when I get back from work.

chenglou:
I've cleaned up the syncing process a little, but haven't re-documented it yet. Feel free to ping me if you're stuck

astrada:
I'm closing this one, and reopening another pull request.

PR #232, opened 2017-11-05
Add comments section

arecvlohe:
If this is an overview, does that include things as simple comments? I tried searching for how to do comments in reason but instead just copy pasted what I found from OCaml docs. Is it implied that I learn OCaml before looking into Reason?


chenglou:
Hello! Comments use `/* here */`. I'll correct it.

It's not totally implied, but OCaml knowledge helps. Since Reason's just a new syntax for it, you can use any existing ocaml tutorial, and use [reason-tools](https://github.com/reasonml/reason-tools) to convert from ocaml syntax to reason syntax on the fly.

Thanks for the PR!

PR #243, opened 2017-11-22
Float exponentiation operator in primitives table

illbexyz:
The float exponentiation operator was missing in the overview documentation.


chenglou:
Thanks!

PR #248, opened 2017-12-01
Add running with Node instructions to quickstart.md

petetnt:
Hi!

I am a total noobie with ReasonML with around 15 minutes of experience under my belt. When going through the quickstart and installing the dependencies everything went smoothly until I wanted to actually try running the code I had written. Took me embarrassingly long to figure out that I can just run the `bs.js` files under `node` to test the output out to see if everything just worked.

I added an extra step of running the demo.bs.js file under `src/` after build with `node` so the user can see at least _some_ output after doing all the work of installing the dependencies and all.

Do tell if one should be running something else altogether, like I said I really have no idea what I am doing yet 👍

### DIFF
--- a/src/pages/guide/javascript/quickstart.md
+++ b/src/pages/guide/javascript/quickstart.md
@@ -15,9 +15,7 @@ cd my-first-app
 npm run start
 ```
 
-It runs in watch mode, so any changes to files will be picked up and compiled.
-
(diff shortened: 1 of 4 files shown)


jamischarles:
Was just about to suggest this same change 👍

chenglou:
Thanks!

PR #252, opened 2017-12-07
Removes funny but unnecessary adjectives

alexUXUI:
I am a fan of editorial voice but think there are better places in the Docs to let the passion and personality come through :)


chenglou:
Thanks!

PR #255, opened 2017-12-09
add re:bench to extra-goodies

glennsl:


chenglou:
Awesome. Added the note that it's compiled using BuckleScript

PR #258, opened 2017-12-11
update Map example to Reason

kimsk:


kimsk:
I just realize that we should not update html file directly (https://github.com/reasonml/reasonml.github.io/pull/154). I will close this one then.

PR #263, opened 2017-12-12
Fix typo on Variant page

skimberk:
Fixed a typo where `type foo = Int(int) | String(int)` should have been `type foo = Int(int) | String(string)` (unless I somehow misinterpreted what's going on).


chenglou:
Thanks!

PR #270, opened 2017-12-20
change words veeery, veeeery to exceptionally and extremely

gnuman:
I get that veeery is used intentionally but it would be good to use simple words.
As a I am not a English speaker, it would be easy to focus on documentation than google words lookup :)


chenglou:
Thanks! This is superseded by #252 already. Btw, our translation efforts are at https://crowdin.com/project/reason

PR #275, opened 2017-12-23
Remove potentially condescending language

canadaduane:
"There, there" can be interpreted in a funny way or a condescending way. Maybe just remove it?


chenglou:
It's a light-hearted reply to a question that's otherwise usually asked rather mean-ly =)

Thanks, but I'd like to keep a light playful tone in an otherwise already long and boring to read docs. Maybe there are things with similar tone

chenglou:
This is stale, but I got another similar feedback, so I rewrote it

PR #277, opened 2017-12-23
Grammar ("Variants")

canadaduane:
Could go with "A variant..." as well, but it seems to work to pluralize it when talking about the forms of variants.


chenglou:
Thanks!

PR #281, opened 2017-12-30
Reason 3 syntax in Array.html

jasonrose:


chenglou:
Damn... did you do this by hand? That page's generated from this: https://github.com/reasonml/reasonml.github.io/tree/800ed72504f8a6dfb7b867f3ccff25f846e9b4bd/generate-docs

jasonrose:
I figured that there was one so I only manually did one page to find out where it was.  I’ll run the script to update everything and update this PR with it tonight.

jasonrose:
I'm abandoning this because the documentation heavily overuses `<code>` blocks for things that are and are not OCaml code.

My original fix tonight was to `refmt`  code blocks as non-interfaces so that inputs like `Array.get a n` would be correctly changed to Reason.  This generates correct text like
```
Array.init(n, f) returns a fresh array of length n, with element number i initialized to the result of f(i). In other terms, Arary.init(n, f) tabulates the results of f applied to the integers 0 to n - 1.
```

However, other code snippets get sugared too much by `refmt` into something not useful, such as the documentation for `Array.get a n`.

On top of that, places that use `<code>` elements that aren't OCaml code, like in Arg.html, become incorrect and broken.  For example, `cmd a b -- c d` becomes `cmd(a, b) -- c(d)`.

chenglou:
Yeah it's hard. Thanks for exploring into it though =)

PR #284, opened 2018-01-02
Replace the index with redex

l4u:
reasonml-community/index is deprecated


chenglou:
The branch has diverged a bit because of me. I've patched this manually in da242a4ca0a20d0653805362a5edb710a5bb3e33. Thanks for the contribution though!

PR #290, opened 2018-01-05
Fix the Exception section's try example

ox:
Currently it renders incorrectly when it's converted from Markdown to HTML; cutting off the matcher. Using the inline code syntax in a Markdown table does not yield the correct syntax, thus a `<table>` has to be used.


chenglou:
Thanks! There's actually a better workaround, done in #291 which also fixes a few other errors, so I'll close this one in favor of it. Thanks for the contribution though! =)

PR #291, opened 2018-01-05
Fix broken formatting in cheatsheet

### LINKED ISSUE #288: Javascript Syntax Cheatsheet: Boolean chart breaking
On the Javascript Syntax Cheatsheet page, the Boolean comparison chart is breaking into multiple extra cells where it should be only 2 cells. 

JS syntax cheatsheet boolean chart break

I forked the repo and found that this chart is breaking on the "Logical OR" operator; this is causing the problem because the operator (1) uses the same Unicode character as the Markdown chart delimiter (2).

OR operator versus Markdown chart delimiter

I played around with it for a minute, and it doesn't appear as though back-slash escaping either `|` character works to fix this problem. Using `&#9474;`/"BOX DRAWINGS LIGHT VERTICAL" could be a solution for displaying two pipe-like characters and be similar in appearance (`││` versus `||`); this is not ideal if the intent is to allow visitors to copy/paste directly from this cheatsheet, as using the alternative vertical line would not be allowed in JS code.

Anyways, just wanted to bring this to your attention. I have [made this change](https://github.com/geoffdavis92/reasonml.github.io/blob/js-syntax-break/docs/syntax-cheatsheet.md#boolean) on my fork if swapping out Unicode characters is desired.

yangshun:
The `|` characters in the [Syntax Cheatsheet's example code](https://reasonml.github.io/docs/en/syntax-cheatsheet.html#boolean) conflict with Markdown's table syntax which results in broken formatting. This PR converts all usage of `|` to `&#124;` and use `<code>` instead of backticks for those cells.

Built the website locally and verified all sections. You can also verify that it renders ok on my [fork](https://github.com/yangshun/reasonml.github.io/blob/cheatsheet/docs/syntax-cheatsheet.md).

**Broken**

screen shot 2018-01-05 at 1 33 48 am

**Fixed**

screen shot 2018-01-05 at 1 35 48 am

Fixes #288

### DIFF
--- a/docs/syntax-cheatsheet.md
+++ b/docs/syntax-cheatsheet.md
@@ -7,90 +7,90 @@ We've worked very hard to make Reason look like JS while preserving OCaml's grea
 
 ## Let Binding
 
-JavaScript                |   Reason
---------------------------|--------------------------------
-`const x = 5;`              |  `let x = 5;`
-`var x = y;`                |  No equivalent (thankfully)
-`let x = 5; x = x + 1;`     |  `let x = ref(5); x := x^ + 1;`
+| JavaScript              | Reason                         |
+| ----------------------- | ------------------------------ |
+| `const x = 5;`          | `let x = 5;`                   |
+| `var x = y;`            | No equivalent (thankfully)     |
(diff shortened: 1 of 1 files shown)


chenglou:
Thanks!

PR #292, opened 2018-01-05
Improvements on links and code rendering

haskellcamargo:
### This pull-request does:

- Fix broken link to `option` type on _Syntax Cheatsheet_
- Bypass Markdown parsing issue for `|`

| Before | Then |
|----|----|
| 2018-01-05-142608_791x425_scrot | 2018-01-05-142622_652x373_scrot |


yangshun:
In the [HTML spec](https://www.w3.org/TR/html5/grouping-content.html#the-pre-element), `<code>` is shown to be within `<pre>`. This [StackOverflow thread](https://stackoverflow.com/a/4611735/1751946) has some examples on this.

I think there are missing styles for `<pre>` and the correct approach would to add the appropriate CSS.

chenglou:
Hey! Sorry, this got a little stale after #291 and fe7d3ea8eb9db3be248def052278b23fd4d452fd. If you'd like, could you submit that `pre` diff separately (if it's still the right way)? For the HOC article, let's hold on it for a bit. Maybe we'll put it on the ReasonReact site instead.

haskellcamargo:
The only change will be the link fix for `option` type. I'll be sending the fix in another PR. Thanks!

PR #298, opened 2018-01-09
Update let-binding.md

Gregoirevda:
Show usage of `and` keyword for defining multiple variables

### DIFF
--- a/docs/let-binding.md
+++ b/docs/let-binding.md
@@ -8,6 +8,7 @@ A "let binding", in other languages, might be called a "variable declaration/ass
 let greeting = "hello!";
 let score = 10;
 let newScore = 10 + score;
+let first = 42 and second = 43;
 ...
 ```
 


Gregoirevda:
Until JS like syntax isn't added (`,`), I still think it's useful to know this exists

jaredly:
I think the only place that `and` is needed is when defining mutually-recursive functions. Could you change the example to show an example of that?

Gregoirevda:
```reason
let a = () => b() and 
b = () => a();
```
I think it's too specific for one of the first pages of documentation 😄 
I was defining
```
reason
let a = "a";
let b = "b";
let c = "c";
```
and didn't know you could do `and`, so thought it could be useful to share

jaredly:
:/ honestly I prefer
```
let a = "a";
let b = "b";
let c = "c";
```
Ideally I'd like reason to throw a syntax error if you're using `and` outside of a `let rec` :P
I think it probably makes sense in ocaml, where there's the awkward `in` that you avoid if you use `and`

jaredly:
another gotcha of `and` -- shadowing semantics that you super won't expect if you're thinking of it as javascript's `let x = 2, y = 3`
```
let a = 1;
let a = 2
and b = a + 3 // b is now 4, not 5
```

kyldvs:
Thanks for the PR, added `and` documentation in a recursion section here: https://reasonml.github.io/docs/en/recursion going to close this out

PR #302, opened 2018-01-11
Add Rung to Reason users

haskellcamargo:
We are also using ReasonML on [Rung](https://app.rung.com.br)!
Rung


chenglou:
Nice =D

PR #309, opened 2018-01-20
add `string_of_int` to fix the access to list item example

arkhamdev:
Fix the example since the type of `myList` is `list(int)`


chenglou:
Hah, thanks!

PR #310, opened 2018-01-20
Add astrocoders as user of Reason!

guilhermedecampo:
Reason is awesome! 
Unfortunately we don't have a svg logo. Hope that one fits.

💯


grsabreu:
svg added

chenglou:
Cool! Thanks =)

PR #315, opened 2018-02-01
Fix typo in chapter about Pattern Matching.

markus-willems:
Should be 'an exception' instead of 'an exceptions'.


chenglou:
Thanks!

PR #316, opened 2018-02-02
Update dead 'converting' link

siggirh:
We do have two different `converting` docs now, `Converting from JS` and `Converting from OCaml` but I updated the doc to link to the dirty one of the two ;)


chenglou:
Thanks!

PR #319, opened 2018-02-09
Added ReasonReact Docker

anirudhmurali:
Added blog post explaining deployment of ReasonReact with a Dockerfile. Let me know if this section is appropriate. TIA.


chenglou:
I've got no expertise in this. But it looks clean, so I'll merge it, thanks =)

PR #322, opened 2018-02-12
update playground compiler to 2.1.0

glennsl:
It's not entirely clear where to put these vendored files since they're in several places, so I put them here and there and it seems to work so...


chenglou:
Thanks! Can we have 2.2.1 instead?

chenglou:
I'm ashamed to say I still don't understand our workflow for `bs.js`. @bobzhang I've published refmt.js when we `npm install reason`. Can we have `bs.js` published too for `npm install bs-platform`? I think it'll simplify lots of meta tasks and tools folks wanna create.

chenglou:
@glennsl Thanks. I've cleaned up the entire process and made syncing a bit less confusing & more automatic. If you don't mind, I'll close this PR now =)

PR #323, opened 2018-02-18
Warn about else branch

yawaramin:
Since leaving out the `else` branch is a potential gotcha.


chenglou:
Thanks! I'll merge but rephrase it a little

PR #324, opened 2018-02-19
minify reason try playground

shrynx:
- modified rollup config to minify try playground file
- built and added minified try version

**Before**
screen shot 2018-02-19 at 14 17 45

**After**
screen shot 2018-02-19 at 14 24 37


shrynx:
not sure if i had to add the minified file to git, 
as it could be integrated as a part of build process but added anyway as it existed in previous git commits.

chenglou:
Thanks! We should check it in, yes

shrynx:
@chenglou oh okay,
Thank for your reply.
anyway i have already checked in the minified try file.
https://github.com/reasonml/reasonml.github.io/pull/324/commits/55e76dd3c3a4705f8b4070f17b557e2b7a4724c7#diff-826f56df0234b1443c4fdf3d6439b4ab

chenglou:
Thanks!

PR #326, opened 2018-03-01
Add note about index callback in iterables

knowbody:


docs/list-and-array.md
```diff
@@ -96,3 +96,5 @@ The above array access/update is just syntax sugar for `Array.get`/`Array.set`.
 ### Tips & Tricks
 
 If you're compiling to JavaScript, know that Reason arrays map straightforwardly to JavaScript arrays, and vice-versa. Thus, even though arrays are fix-sized on native, you can still use the `Js.Array` API to resize them. This is fine.
+
+If you need to map through the `Js.Array` and get an index in the callback use `.mapi` instead of `.map` as `.map` does not provide the index.
```
malisbad: It doesn't have to be a `Js.Array`. Any arrays in Reason have the `.map()` and `.mapi()` difference. Same with `.iter()` and `.iteri()`.

https://reasonml.github.io/api/Array.html

knowbody: ok should I change it to something like:

> If you want to be provided with an index in the callback once you use any methods that iterate over an Array, add `i` into the method name, i.e. `.mapi()` instead of `.map()` or `.iteri()` instead of `iter()`

malisbad: Personally, I would note that if you're coming from JavaScript, that map behaves differently.  To get the more familiar callback signature of `(index, element) => {}`, you would use `.mapi()`. If you just need `element` in the callback, use `.map()`. and maybe link to the API reference?

knowbody: fixed

docs/list-and-array.md
```diff
@@ -96,3 +96,5 @@ The above array access/update is just syntax sugar for `Array.get`/`Array.set`.
 ### Tips & Tricks
 
 If you're compiling to JavaScript, know that Reason arrays map straightforwardly to JavaScript arrays, and vice-versa. Thus, even though arrays are fix-sized on native, you can still use the `Js.Array` API to resize them. This is fine.
+
+If you're coming from JavaScript, iterables behave differently. To get the more familiar callback signature of `(element, index) => {}`, you would use `.mapi()` or `.iteri()`. If you just need an element in the callback, use `.map()`. See the [API reference](https://reasonml.github.io/api/Array.html#VALmapi).
```
malisbad: Javascript's `.map()`'s callback has an argument signature of `(element[, index,[array]])`. Reason's `.mapi()` callback has an argument signature of `(index, element)`. It's more familiar to the JS, but the order is reversed, and there is no array.

kevinsimper: @malisbad Good point

PR #328, opened 2018-03-08
rename physicalEquality to structuralEquality

nikgraf:
Here an Issue with a longer explanation: https://github.com/facebook/reason/issues/1851


chenglou:
cc @jordwalke @hcarty lol

jordwalke:
Yes, thank you.

PR #332, opened 2018-03-15
Add ReasonReact support to the playground

astrada:
* Update BuckleScript compiler (bs.js) to 2.2.3
* Add PPX rewriter jsxV2.js (jsoo_reactjs_jsx_ppx_v2.js from BuckleScript/2.2.3)
* Update setupSomeArtifacts.js to build all the new artifacts
* Add new dependencies:
    - reason-react: to generate bundles for loading ReasonReact modules into the playground
    - bs-platform: to compile reason-react
    - react-async-script-loader: to dynamically load new assets
* Add Preview component (with cleanslate to reset css)


astrada:
I think I managed rebasing my ReasonReact patch (#229). It probably requires a bit of testing (I tested it on Chrome and Firefox on Ubuntu).

chenglou:
Alright, I'm merging this! There are a few tweaks and bug fixes needed for the playground, but I'll apply them afterward. I haven't tried to repro the syncing step; this will need to be one-click. Might need to ping you later.

Thanks so much!

PR #334, opened 2018-03-20
Add a copy/paste friendly example to promise docs

malisbad:
Original example left out a copy/paste friendly example to play with in the playground.


chenglou:
Thanks!

PR #335, opened 2018-03-21
api/List.html correct typo for map def

jeanbaptisteassouad:


jeanbaptisteassouad:
my bad, their is no error, because ('a, 'b) => 'c is the same as 'a => 'b => 'c due to Reason syntactic sugar.
I was reading ('a, 'b) as a pair.

zploskey:
Just for future reference, the html files are generated from from the doc comments in the code. Any changes to content should be made there.

PR #338, opened 2018-03-23
Fix wrong `reason-cli` url

andrepolischuk:
`https://www.npmjs.com/package/reason-cli/` with end slash returns 404


chenglou:
Hi! Really? It seem to work here?

zploskey:
It's strange, I got the 404 when this was first posted but it seems to work now. I wonder if something changed on the NPM site.

andrepolischuk:
It seems to fixed on the npmjs site. Sorry.

chenglou:
Np! I'll close this then. Thanks!

PR #347, opened 2018-04-03
Update Faq with link to ocaml manual

kevinsimper:
As a beginner I was confused and was looking for a function to convert a int to string, then found this answer but then had no idea where to find the "standard library". It would be quicker to link to a trusted source to read about the standard library for beginners and see other functions as well.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -23,6 +23,10 @@ See [here](quickstart-javascript.md). Reason's a syntax for OCaml and supports a
 ### Where do all these `print_endline`, `string_of_int` functions come from?
 They're from the standard library, pre-`open`ed during the compilation of your file. This is why you see them in scope.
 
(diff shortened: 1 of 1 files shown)


chenglou:
Hello! The link is the API page. Was that unclear?

kevinsimper:
Which link? There were none already?

cem2ran:
@kevinsimper Cheng is referring to the API link in the website header menu. This leads to: https://reasonml.github.io/api/index.html wherein you can find the Pervasives module. I don't think it would hurt to include that link.

kevinsimper:
I updated the link to point to the api section instead 😄

@cem2ran thanks a lot! 👍

jaredly:
awesome, sorry for the delay!

kevinsimper:
Thanks @jaredly 😄

PR #353, opened 2018-04-12
Update small typo in FAQ :)

JasoonS:


jaredly:
`_ => 1` will actually fail to parse iirc. I think we need `(_) => 1`

JasoonS:
Thanks

PR #360, opened 2018-04-23
Replace Js.true_ with just true etc.

mmiszy:
See: https://bucklescript.github.io/blog/2018/04/16/release-3-0-0.html


jaredly:
Yayyyy

PR #361, opened 2018-04-27
Add Dernier Cri to Reason Users

dusterherz:
Hi !
We started to use actively Reason-React and Reason-React-Native in our projects. We've also started to work on a port of the react-table library ! We will be happy to be added in the users of reason.


jaredly:
sweet :)

PR #363, opened 2018-05-02
Remove comma in first example.

bokuo-okubo:
Hi.
In 27fd7ceb89cd6041557aecf7f0ef115d9b81a824, it seems remove the `%` at the first example,
but if we use single `%`, it should contained as an expression.


jaredly:
oh good call

PR #364, opened 2018-05-03
wrap ellipsises with comment syntax

flppv:
Hi!
I think it will look more consistent in docs, if dots would be wrapped with comment syntax


jaredly:
solid, thanks!

PR #367, opened 2018-05-13
Adding 'install with yarn' to troubleshooting

jondot:
Hi,
Just finished a two-hour session trying to figure out why Reason CLI tools won't install.
Apparently it choked on a path which it thought `is too deep inside filesystem, Esy won't be able to relocate binaries`.
This specific error was also silent, so I had to go into the esyBuildRelease shell script and make it spit it out.

This [issue](https://github.com/reasonml/reason-cli/issues/45) hinted using Yarn would do off with all that, and it did, installing quickly and happily. 

Might also be worth recommending Yarn as a default running for installing over npm, but I will leave that to more knowledgeable documentation editor.

### DIFF
--- a/docs/global-installation.md
+++ b/docs/global-installation.md
@@ -32,6 +32,12 @@ opam install reason.3.0.4 merlin.2.5.4
 
 If your installation fails, it might be because you're on npm 5.4.0 (`npm --version`). There was a known bug in npm that's fixed in 5.4.2. Upgrade `npm` and things should work.
 
+If _that_ fails, try using Yarn instead of npm (Yarn hides deep-path issues from OCaml):
+
+```
+yarn global add <package>
+```
+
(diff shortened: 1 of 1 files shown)


chenglou:
cc @andreypopp

andreypopp:
I'm not sure that yarn would alway solve this issue — for me it looks like it just happened that yarn installs into a fs location which isn't so deep as the one used by npm.

The issue with swallowing the actual error is the real one — I've pushed a fix into esy/esy#182.

jondot:
Thanks. I can confirm that the difference is where Yarn installs vs npm. But since the documentation states "if this doesn't work, try X, and if not, try Y, ..." and so on, this becomes just another solution to  try.

kevinsimper:
I think this pull request can be closed as `yarn` is now used in the docs. cc. @chenglou

chenglou:
You're right. Sorry for not merging this earlier and thanks again for the contribution @jondot.

PR #368, opened 2018-05-15
We change company name

cullophid:
We have changed name from Sentia(https://sentia.ai) to mesopo(https://mesopo.ai). 
I have replace the logo on the list of companies using reason.
I have also removed gain.ai. It was the first product that we used reasonml on, but it is now discontinued. (I also dont think we should have two entries on the list, given all the great companies that are using reasonml)


jaredly:
coolio

PR #370, opened 2018-05-17
add video course to articles & videos section

nikgraf:
Not sure if it's the best place here. Let me know if I should change it or you have concern adding it at all.


chenglou:
It's fine putting it here! My concern is that it might not be up-to-date in a few months

nikgraf:
@chenglou my ambition is to keep it up to date, but happy to remove it from here in case that doesn't happen

chenglou:
Cool. Either way, thanks for the work, as always =)

PR #372, opened 2018-05-18
Update to reflect current setup

ostera:
Just now I tried it out and the path changed!


chenglou:
The paths should be right; did you change `package-specs` in bsconfig? Or do you have a very old bs-platform?

chenglou:
Ping =)

ostera:
Checking again!

ostera:
Confirmed to no longer be relevant 🆗 Closing!

PR #378, opened 2018-05-22
Fix small typo

vegansven:
{name: "Bob", Job: Pro**g**ammer({fullTime: Yes | Maybe})} => ...
{name: "Bob", Job: Pro**gr**ammer({fullTime: Yes | Maybe})} => ...


chenglou:
Thanks!

Issue #377, opened 2018-05-22
Global Install instructions need a remark


Kakadu:
https://reasonml.github.io/docs/en/global-installation.html

I was trying to use `npm install -g reason-cli` to get everything on my GNU/Linux machine but it downloaded MacOS tarballs. Can you explicitly mention that `@3.1.0-linux` is really required?

chenglou:
Can be closed now

PR #379, opened 2018-05-22
Clarification about Darwin by default

Kakadu:
A clarification what will happen if we run `npm install -g reason-cli` without version

See issue #377

### DIFF
--- a/docs/global-installation.md
+++ b/docs/global-installation.md
@@ -12,7 +12,7 @@ title: Global Installation
 | **Linux**   | `npm install -g reason-cli@3.1.0-linux`
 | **Windows** | Please see https://github.com/reasonml/reasonml.github.io/issues/195
 
-Note 1: Running `npm install -g reason-cli` without version specification will install Darwin executables on *any* host system. This bug is not yet fixed but will be fixed in near future.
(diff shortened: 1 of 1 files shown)


docs/global-installation.md
```diff
@@ -8,11 +8,13 @@ title: Global Installation
 
 | Platform  | Install command
 |-----------|-------------------------------------------------------------------------------------------------
-| **macOS**     | `npm install -g reason-cli@3.1.0-darwin`
+| **macOS**     | `npm install -g reason-cli` or `npm install -g reason-cli@3.1.0-darwin`
```
andreypopp: I don't think we should advertise `npm install -g reason-cli` but instead fix `latest` release tag not to point to Darwin build.

jaredly: what would latest point to?

andreypopp: Either:
- a dummy package which errors with a message to install a correct package
- a dummy package which installs the correct package during a postinstall phase (cc @jordwalke why we didn't do this, I can't remember)

kevinsimper:
This can be closed now since the reason-cli is not mentioned anymore :)

cc. @chenglou

chenglou:
Thanks

PR #384, opened 2018-05-25
Fix reference to Reason objects

rparrapy:
Currently, the previous section only has a link to Bucklescript's Object record mode, not really Reason objects.


chenglou:
Thanks! I've changed the link to use the markdown file (docusaurus compiles that to the right html link)

PR #389, opened 2018-06-05
Update the "full build" docs based on my experiences

phated:
As per our back-and-forth on twitter, I believe this is the full set of docs for building the playground from scratch.  It's what I did.

Please note that bucklescript's `repl.js` no longer references AMD so that step was completely removed.

Aside: I plan to submit some changes to repl.js that make these docs simpler but not sure when I'll have time.


chenglou:
Thanks!

PR #392, opened 2018-06-08
search right language in search bar

### LINKED ISSUE #388: Search right language in search bar
It seems that reasonml website search bar doesn't search the correct language. It only search english documentation when I'm actually on Espanol

wrong

I saw @chenglou restricting the doc search to 'english' only based on https://github.com/algolia/docsearch-configs/pull/292 because searching 'foo' gives 5+ result across all languages. Till today https://github.com/algolia/docsearch-configs/blob/master/configs/reason.json is still restricted to english

Some suggestion.
You can search the right language by reverting that commit and change this
https://github.com/reasonml/reasonml.github.io/blob/95f4d390b50f823a7b88f21098fec30bc34c6162/website/siteConfig.js#L265-L268

to
```js
  algolia: {
    apiKey: "966d1e412f67114a07dc0afe44b19b53",
    indexName: "reason",
    facetFilters: ["lang:LANGUAGE"]
  },
``` 

P.S : I can help to submit a PR if you want

This is because `LANGUAGE` will be replaced by Docusaurus to the right language. 
Source code:
https://github.com/facebook/Docusaurus/blob/d28b864a59fabeea45add8c090a13de7d0530de5/lib/core/Site.js#L128-L142
```js
algoliaOptions: ${JSON.stringify(
   this.props.config.algolia.algoliaOptions
)
 .replace('VERSION', docsVersion)
 .replace('LANGUAGE', this.props.language)}
```

Example on my site:
Only search for 'korean' documentation when on 'korean'
dynamic language

Only search for 'chinese' documentation when on 'chinese'
cn

cc @joelmarcey @yangshun . Maybe we should update Docusaurus documentation on search bar. I don't think user know this extra processing that we did. We only refer them to https://community.algolia.com/docsearch/documentation/ 😢 based on https://docusaurus.io/docs/en/search.html#enabling-the-search-bar

endiliey:
## Motivation

Fixes #388 

## Notes
Only will work after https://github.com/algolia/docsearch-configs/pull/443 is merged

## Test Plan
```
cd website
yarn start
```

Choose different language, type in the search bar any keyword, should search only that language

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -269,7 +269,10 @@ const siteConfig = {
   },
   algolia: {
     apiKey: "966d1e412f67114a07dc0afe44b19b53",
-    indexName: "reason"
+    indexName: "reason",
+    algoliaOptions: {
+      facetFilters: ["lang:LANGUAGE"] 
+    }
   },
   disableHeaderTitle: true,
 };


JoelMarcey:
@chenglou Can you merge this?

chenglou:
Thanks!

endiliey:
Works like magic
chinese

chenglou:
Thanks for looking into this!

We cam activate it for BuckleScript and ReasonReact too I guess

JoelMarcey:
@endiliey Awesome!

We should probably do the same for Docusaurus, now that we have enabled translations 😄

PR #393, opened 2018-06-10
Add a simple example of creating a reason object

arecvlohe:
Maybe I am just slow but when I first looked at the docs on objects I was stumped on how to create a simple object. After fooling around with it for about 20 min I finally figured out how they work. Coming from a JS background, objects don't have the idea of public and private methods in an explicit sense. In constructors, yes, but in plain ole' objects no, which is what I think of when I think of objects. I added a simple example, and referenced the other as advanced. I think the example I added shows how to create an object in its most simplest form while also referencing the very first example at the start of the doc. I think this would help newcomers, and advanced developers, to recognize that an object behaves much more like a constructor than an object, in the JS sense, and reduce confusion going forward in the doc.

### DIFF
--- a/docs/object.md
+++ b/docs/object.md
@@ -32,6 +32,25 @@ Two dots, also called an elision, indicate that this is an "open" object type, a
 
 ### Creation
 
+## Simple
+
+```reason
+type tesla = {
+  .
+  color: string,
+};
+
+let obj: tesla = {
+  pub color = "Red"
+};
+
+Js.log(obj#color) /* "Red" */
+```
+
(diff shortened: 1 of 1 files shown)


jaredly:
Yeah, I like where this is going. It's made a little complicated by the fact that the `property` you describe is in fact a function w/ zero arguments. Could you adjust the language to reflect that?
Objects only export methods, and all properties are private.

arecvlohe:
@jaredly I updated the doc to reflect what you are saying. I wasn't sure if `val`'s are what you mean by properties. Let me know if this is not accurate.

jaredly:
fantastic! this looks much clearer

PR #394, opened 2018-06-11
Update function.md

Gregoirevda:
Delete `with default value` in comment: There was no default value
add `unit` in type signature


chenglou:
oops, thanks

PR #395, opened 2018-06-15
Adds more clarity to download progress note

kyleshevlin:
I believe the use of `but` was incorrect. It almost read as a double negative. I added more clarity to the section. I certainly thought it was hanging the first few times before I read the note more closely!


jaredly:
thanks :)

PR #398, opened 2018-06-24
Update Docusaurus to 1.3.1

yangshun:
## Summary

Docusaurus 1.3.1 brings about some [backwards-incompatible CSS changes](https://github.com/facebook/Docusaurus/blob/master/CHANGELOG.md#breaking-changes). So the team is helping the major users update 😄 

I removed some custom CSS pertaining to the side nav as the fixed nav is now part of Docusaurus core. Also changed the code blocks in table markup and standardized them across the site.

cc @JoelMarcey @endiliey 

## Test plan

Ran the site locally and visited every page.

screen shot 2018-06-23 at 10 22 21 pm
screen shot 2018-06-23 at 10 22 28 pm

### DIFF
--- a/docs/community.md
+++ b/docs/community.md
@@ -12,7 +12,7 @@ Come say hi!
 - [Reddit](https://www.reddit.com/r/reasonml/)
 - [Stack Overflow](http://stackoverflow.com/questions/tagged/reason)
 
-## Github
+## GitHub
 
 - [Reason Org](https://github.com/reasonml)
 - [Reason Community Org](https://github.com/reasonml-community)
--- a/docs/comparison-to-ocaml.md
+++ b/docs/comparison-to-ocaml.md
@@ -29,26 +29,31 @@ Reference | `x != y` | `x !== y`
 Reason's lexical scoping is exactly the same as OCaml's, but let bindings syntactically resemble "block scope" which is more familiar to many developers. In Reason, they are created with `{}` braces, which may contain both `let` bindings and imperative commands, separated by `;`. All blocks evaluate to the last line and the semicolon on the last line is optional. `{}` braces are only needed if you have more than one item to chain together via `;`.
 
 <table>
-  <thead><tr> <th><p>OCaml</p></th><th><p>Reason</p></th></tr></thead>
+  <thead>
+    <tr>
+      <th>OCaml</th>
+      <th>Reason</th>
+    </tr>
+  </thead>
+  <tbody>
   <tr>
     <td>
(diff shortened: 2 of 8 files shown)


chenglou:
That's some kind of open source support... thank you so much!

- I've noticed that the cut off point for shrinking into a navigation-at-the-top view is now 1024px. However, the Edit button for each page still stays until 736px. Is this intentional? Also, the page header's cut off point is still 736px.
- The Edit button isn't vertically centered relative to the title, and has extra margin to the right (misaligned vs the content area and the button at the bottom for navigating to the next page):
  screen shot 2018-06-28 at 12 46 15 am
-  Search style broke on small view:
screen shot 2018-06-28 at 12 47 54 am
- We have our custom code style where the code border horizontally extends beyond the content area. The new cut-off point of 1024px makes the code block look a bit weird from 736px to 1024px. This is our own hack; we'll fix it.
- Table view indeed looks much nicer now =D
- You fixed the blockquote first paragraph's padding! Thanks
- Headers are now black and bold; is this intentional? I'm fine either way

yangshun:
> I've noticed that the cut off point for shrinking into a navigation-at-the-top view is now 1024px. However, the Edit button for each page still stays until 736px. Is this intentional? 

Not really, thanks for pointing it out.

> Also, the page header's cut off point is still 736px.

> The Edit button isn't vertically centered relative to the title, and has extra margin to the right (misaligned vs the content area and the button at the bottom for navigating to the next page):

Will fix that!

> Search style broke on small view:

Will fix that!

> We have our custom code style where the code border horizontally extends beyond the content area. The new cut-off point of 1024px makes the code block look a bit weird from 736px to 1024px. This is our own hack; we'll fix it. 

Got it. I think with the new styling the code border looks better without extending beyond the content.

> Table view indeed looks much nicer now =D 

Yay!

> You fixed the blockquote first paragraph's padding! Thanks 

That has been annoying me for a while. Glad that it's fixed now.

> Headers are now black and bold; is this intentional?

Yep I think it's more professional for documentation websites to have black headings. I referenced a ton of existing documentation sites and they all have neutral color headings. Users could easily override this if they wanted, so feel free 😄 

I'll address the above-mentioned CSS issues in Docusaurus and we'll release a 1.3.1 and I'll update this PR with 1.3.1. Thanks for the thorough review!

chenglou:
Thanks so much =)
Merged and filed an extra PR at https://github.com/facebook/Docusaurus/pull/817

PR #402, opened 2018-07-06
Replace confusing ReasonReact instructions

jsdf:
If people try to follow the instructions on the Quickstart page, and then enter the command shown to use the 'react' theme for bsb -init, they will encounter a compiler error due to the fact that npm install hasn't been run. This is not a great first experience.

Additionally, the instructions on this page describe running the code under node, which really makes no sense for ReasonReact. We'd really be better off just having them go to the ReasonReact instructions instead.


chenglou:
Hey James! =D

PR #403, opened 2018-07-11
add command for using yarn

vonwao:
yarn syntax differs, make it easy to install


chenglou:
Thanks! Actually, we'll default to yarn

PR #406, opened 2018-07-16
Remove dead link in integer-and-float.md

chucksmash:
http://www.lexifi.com/blog/unboxed-floats-ocaml now 404s, unfortunately.


chenglou:
Oops sorry, I've already merged #408 just now. Thanks a lot for the contribution though! =)

PR #414, opened 2018-08-04
Update quickstart-javascript.md

JosephJNK:
The default behavior of the quickstart project is to put compiled files alongside source files and includes them in git, which is very unusual for a compile-to-JS language, and goes against the expectations/best practices of developers who are used to other such languages. [This post](https://reasonml.chat/t/should-i-include-or-exclude-bs-js-files-in-git/60) makes it sound like this is a common source of confusion. It would likely be less friction for such developers if the behavior of the quickstart project was to put compiled files in `lib` by default, but short of this an explanation of how to change this behavior would be helpful.

### DIFF
--- a/docs/quickstart-javascript.md
+++ b/docs/quickstart-javascript.md
@@ -21,6 +21,8 @@ We use the BuckleScript compiler to compile Reason to JavaScript, then we use No
 
 When developing, instead of running `npm run build` each time, run `npm run start` to start a watcher that recompiles on file change.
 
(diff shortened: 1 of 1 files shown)


chenglou:
Thanks! Yeah, the explanation is inside [BuckleScript doc page](https://bucklescript.github.io/docs/en/build-overview#tips-tricks), but since it seems that folks don't notice the BS docs much, a bit of redundancy here is nice.

PR #421, opened 2018-08-14
Add docs about fast-pipe

kevinsimper:


chenglou:
Hah, Thanks for getting to this =). We'll adjust the wording a little, but overall this is saving us some work!

PR #422, opened 2018-08-14
Clarify use of ~ in labeled arguments

joefiorini:
A developer on my team experienced some confusion with the `~` symbol in front of labeled arguments. I thought it might make things clearer to explicitly point out its usage.


chenglou:
Thanks!

PR #424, opened 2018-08-16
Update fast-pipe.md

kevinsimper:
When I first read about fast pipe, the thing where it get put as the first argument even if there is multiple, was not straight forward as it was something I had not seen before.

In this PR I just make the section a little bit more expanded to not just be a byline.


chenglou:
Thanks!

PR #429, opened 2018-08-24
Add padding to API docs

benjamminj:
I noticed that with the contents of the API docs pushed up against the edge of screen, it can get a little difficult to read at times. Adding a little padding makes things a little easier on the eyes ✨ 

Here's what it looks like with the extra padding.

screen shot 2018-08-23 at 6 36 23 pm

### DIFF
--- a/website/static/css/api.css
+++ b/website/static/css/api.css
@@ -10,7 +10,7 @@
 }
 
 .ocamldoc {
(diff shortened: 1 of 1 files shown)


kevinsimper:
@benjamminj It is good 👍

chenglou:
Thanks

PR #430, opened 2018-08-25
Add placeholder docs to fast-pipe

rickyvetter:
Some discussion with @danielo515 in discord about how docs here would be nice. I added them to fast-pipe specifically because it's not clear that using them widely outside of pipes is the right idea. Open to edits/improvements if anyone sees anything.

### DIFF
--- a/docs/fast-pipe.md
+++ b/docs/fast-pipe.md
@@ -106,6 +106,13 @@ let result = Some(preprocess(name))
 
 ## Pipe Placeholders
 
+A placeholder is written as an underscore and it tells Reason that you want to fill in an argument of a function later. These two have equivalent meaning:
+
+```reason
+let addTo7 = (x) => add3(3, x, 4);
+let addTo7 = add3(3, _, 4);
+```
+
 Sometimes you don't want to pipe the value you have into the first position. In these cases you can mark a placeholder value to show which argument you would like to pipe into.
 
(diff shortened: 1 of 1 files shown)


Daniel15:
I don't think I'm in the Reason Discord... You might be thinking of someone
else :P

danielo515:
Actually I am danielo515, exactly that's my name so it is correct.

docs/fast-pipe.md
```diff
@@ -103,3 +103,28 @@ We turn this into:
...
+Sometimes you don't want to pipe the value you have into the first position. In these cases you can mark a placeholder value to show which argument you would like to pipe into.
+
+Let's say you have a function `namePerson`, which takes a `person` then a `name` argument. If you are transforming a person then pipe will work as-is:
+
+```reason
+makePerson(~age=47, ())
+  ->namePerson("Jane");
+```
+
+But if you have a name that you want to apply to a person object, you can use a placeholder:
```
kevinsimper: You should show what a placeholder is first and then show how to use it later, then it will become more clear. It will also make it more clear if you don't start the sentence with "But" here in this case when explaining a new concept.

docs/fast-pipe.md
```diff
@@ -106,6 +106,13 @@ let result = Some(preprocess(name))
 
 ## Pipe Placeholders
 
+A placeholder is written as an underscore and it tells Reason that you want to fill in an argument of a function later. These two have equivalent meaning:
```
kevinsimper: Awesome 👍

PR #431, opened 2018-08-26
Update cheatsheet with map/forEach examples

kevinsimper:
This is good for a beginner to see how ReasonML is handling looping over an array. 

How the array is giving as a argument.

### DIFF
--- a/docs/syntax-cheatsheet.md
+++ b/docs/syntax-cheatsheet.md
@@ -149,6 +149,13 @@ Both JavaScript and Reason support currying, but Reason currying is **built-in a
 | `for (let i = 10; i >= 0; i--) {...}` | `for (i in 10 downto 0) {...}` |
 | `while (true) {...}`                  | Same                           |
 
+## forEach / map
+
+| JavaScript                               | Reason                                    |
+|------------------------------------------|-------------------------------------------|
+|`[1,2,3].forEach(i => { console.log(i) })`|`Array.iter(i => { Js.log(i) }, [|1,2,3|])`|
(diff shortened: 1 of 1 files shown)


kevinsimper:
Hi @rickyvetter, what do you think of this PR? :)

docs/syntax-cheatsheet.md
```diff
@@ -149,6 +149,13 @@ Both JavaScript and Reason support currying, but Reason currying is **built-in a
 | `for (let i = 10; i >= 0; i--) {...}` | `for (i in 10 downto 0) {...}` |
 | `while (true) {...}`                  | Same                           |
 
+## forEach / map
+
+| JavaScript                               | Reason                                    |
+|------------------------------------------|-------------------------------------------|
+|`[1,2,3].forEach(i => { console.log(i) })`|`Array.iter(i => { Js.log(i) }, [|1,2,3|])`|
```
yawaramin: Reason version can be `Array.iter(Js.log, [|1, 2, 3|])`

docs/syntax-cheatsheet.md
```diff
@@ -149,6 +149,13 @@ Both JavaScript and Reason support currying, but Reason currying is **built-in a
...
+|`[1,2,3].map(i => { return i + 1 })`      |`Array.map(i => i + 1, [|1,2,3|])`         |
```
yawaramin: Reason version can be `Array.map(succ, [|1, 2, 3|])`. Admittedly `i => i + 1` is just as valid but if we want to show the idiomatic way of doing things in OCaml/Reason we might as well teach the built-in functions...

PR #439, opened 2018-09-14
module.md: Clarify extending module signatures

zploskey:
The extending module signatures example does not compile unless a
concrete module `BaseComponent` is defined. Although these modules were
defined earlier in this document, they are now a bit separated and this
has caused some confusion for users trying to compile just this part.

This modifies the example to note that a normal `include` will work for
extending another module type and provides a separate example of using
`include (module type of ModuleName)` to extend a module that doesn't
have a defined module type.

There was a dicussion of this on the Discord which prompted an attempt
to improve this section.

### DIFF
--- a/docs/module.md
+++ b/docs/module.md
@@ -208,7 +208,7 @@ exposing the variant, it'll only tell you "it's `Company.profession`".
 
 ### Extending module signatures
 
-Like modules themselves, module signatures can also be extended through `include module type of Foo`:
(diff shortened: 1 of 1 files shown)


kevinsimper:
I think you added explanation is good 👍

zploskey:
Thanks, I think this is clearer. Also I edited the PR text to not have all these awkward line breaks. You would think GitHub might strip them when they generate the initial PR text from a commit but not yet I guess!

PR #441, opened 2018-09-15
Add note about reason-cli tools

jsdf:
It seems to me that these editor integrations won't work without reason-cli tools. I don't see anything in the current install instructions which includes them.


chenglou:
@jsdf you no longer need that actually! Anything not working for you?

PR #442, opened 2018-09-15
Add LaTeX Base (https://latexbase.com) to users list

lambdabaa:
And thank you all for your awesome work on the reason ecosystem! We've had a great experience with the language and tooling so far.


chenglou:
Thanks! You're welcome =)

PR #448, opened 2018-10-10
Vertically aligning superscript

### LINKED ISSUE #433: Superscript appears as normal text
e.g. on https://reasonml.github.io/api/Random.html, 2<sup>30</sup> appears as 230.

This seems to be due to using a CSS reset without then adding back style for the `sup` element.

A style like this should be added:

```css
sup {
    vertical-align: super;
    font-size: 80%;
}
```

danielterwiel:
Fixes #433 

Changes `sup` tags inside the documentation that render like [this](https://reasonml.github.io/api/Int32.html#VALmax_int):

> let max_int: int32;
> 
> The greatest representable 32-bit integer, 231 - 1.

To:

> let max_int: int32;
> 
> The greatest representable 32-bit integer, 2<sup>31 - 1</sup>.

### DIFF
--- a/website/static/css/api.css
+++ b/website/static/css/api.css
@@ -220,3 +220,7 @@
   font-family: Helvetica Neue, Helvetica, Arial, Lucida Grande, sans-serif;
   padding-bottom: 16px;
 }
+sup.superscript {
+  vertical-align: super;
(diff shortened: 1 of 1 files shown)


chenglou:
Thanks!

PR #449, opened 2018-10-13
Should the declaration not be mutable?

naartjie:
Looking at this, it seems to be the case: https://reasonml.github.io/api/Pervasives.html#6_References

The desugared version has been documented like this for a long time, so I might be missing something bigger going on here, that's why I posted this as a question ;-)


naartjie:
Gotcha, thanks for replying. Sorry for getting it wrong

chenglou:
Np!

PR #452, opened 2018-10-17
[Doc] Fixed quotes for strings

kurideja:
Fixed code examples. Single quotes are for chars, and in documentation they were used for strings.


naartjie:
I think it was done intentionally, since that particular source is JS.

kurideja:
Oh my...

PR #459, opened 2018-11-07
Fix external filter function type

kana-sama:


enieber:
@kana-sama please resolve conflit

PR #460, opened 2018-11-07
Fix record field typo

kana-sama:


kyldvs:
Thanks for the PR, these pattern matching docs were recently updated so I'll close this out

PR #465, opened 2018-11-10
Add link to to rtop (reason-cli)

nielsbom:
A local REPL is very useful as well.


nielsbom:
I think that's a personal record wrt oldest PR merged.

PR #467, opened 2018-11-10
Clear up prose on optional labeled arguments

nielsbom:
Tried to make the prose a little clearer.


jordwalke:
Thank you. The phrasing sounds like an improvement.

PR #478, opened 2018-11-26
Add xorlab to users

zenokoller:


zenokoller:
Thank you! Yes, are we the only (official) Swiss ReasonML user so far?

Sent from my mobile

cristianoc:
I think that's true from the current logos.

PR #485, opened 2018-12-08
Change name from `Fast Pipe` to `Pipe First`

nikgraf:
I'm aware this is a radical change and this PR can be a place for discussion first.

When teaching Reason people asked: Why is it called `fast pipe`? In addition they regularly find the pipe operator and then get really confused that they are not doing the same.

My suggestion would be  to establish a consistent naming theme for both pipe operators:
- Pipe first `->`
- Pipe last `|>`

Advantage of this convention is that it's very easy to use in the natural language when teaching.

This partly based on a discussion on the #bikeshedding thread in Discord. Thanks to @ryyppy for bringing it up 👍 /cc @jordwalke

### DIFF
--- a/docs/fast-pipe.md
+++ b/docs/fast-pipe.md
@@ -1,5 +1,5 @@
 ---
-title: Fast Pipe
+title: Pipe First
 ---
 
 `->` is a convenient operator that allows you to "flip" your code inside-out. `a(b)` becomes `b->a`. It's a piece of syntax that doesn't have any runtime cost.


anmonteiro:
thanks for jump starting this proposal. I agree with renaming it, but I'd go even further and revamp those docs too
> -> is a convenient operator that allows you to "flip" your code inside-out. a(b) becomes b->a.

wouldn't this feel clearer if we mentioned somewhere that -> applies a with b as its first argument?

> It's a piece of syntax that doesn't have any runtime cost.

Additionally, this isn't true if you define it in your native code. Only if using the PPX (which isn't included in native by default).
Adding a note about needing to use https://github.com/IwanKaramazow/FastPipe in native would also be helpful IMO

jordwalke:
I'll take a look at further cleaning up this doc, including @anmonteiro's suggestions. For now, First Pipe is officially part of the Reason _concrete_ syntax, but its meaning is able to be defined via ppx functions and BuckleScript is the thing that defines the First Pipe's meaning. That should be noted. For now, I like this PR. Let's see if anyone chimes in with an objection.
Multiple people have expressed confusion about the name "fast pipe" - thinking this has something to do with performance. First pipe is way more clear in my opinion.

nikgraf:
One note: I renamed it to "Pipe First". I asked a few people directly and they prefered "First Pipe" (also what @jordwalke used)

I can update the PR if we agree on one

jordwalke:
Yeah, first pipe makes more sense because it's a noun. You can talk about "first pipe" like it's a think that exists.
"Pipe first" is a thing you do, not a noun.

jordwalke:
Please update to `First Pipe`

mlms13:
I like the names "Pipe First" and "Pipe Last" (as opposed to "First Pipe", "Last Pipe"). It lines up with Clojure's naming of thread-first and thread-last (which I assume was at least some of the inspiration for `->`). It also seems like a more natural answer to the question "what does this operator do".

But either of those options seems like an improvement over "Fast" pipe. :)

jordwalke:
Okay thank you all for the feedback everyone. Going to merge this one.

PR #487, opened 2018-12-17
Update Ast_mapper.html

arbipher:
The type of `(Const_int 42)` is `Asttypes.constant`, but the type required here is `Parsetree.constant`


arbipher:
I modified the demo code to pass the type check.

PR #491, opened 2019-01-04
more example of `all`, `all2` and `race`, and link to async/await proposal

truongsinh:


alavkx:
The link to the async/await discussion can be safely added. Do we still want to push `Js.Promise` as the best way to use promises instead of the alternative API's like https://github.com/aantron/promise?

PR #526, opened 2019-03-31
Update Reason local scope example

yawaramin:
To exactly match the semantics implied by the OCaml syntax, i.e. the visibility of `ten`.


yawaramin:
As @jordwalke [pointed out](https://twitter.com/jordwalke/status/1112504821917990914), this is not exactly equivalent to the OCaml on the left (which means something different if there's a semicolon-sequencing and a following expression). Therefore, closing this PR.

PR #529, opened 2019-04-17
Use published Docusaurus (fixes #520)

### LINKED ISSUE #520: Set page titles that relate to content
The website has the same very long title for all pages. This is unhelpful when scanning browser history.

image

phated:
The patch @chenglou introduced into docusaurus for the Reason syntax highlighting was upstreamed as https://github.com/facebook/Docusaurus/pull/1016

This also fixes #520 due to an upstream fix.


chenglou:
Thanks!

PR #530, opened 2019-04-17
Remove ReasonConf advertisement

phated:
Since the conf is over, it felt weird to see that on the homepage.  I kept the promo `div` wrapper so other things can be advertised later (ReasonConf US? 😉)


phated:
This was done in #536

PR #534, opened 2019-04-19
update link to currying example

shaver:
The section on currying links to a playground example with an outdated URL. This just updates it to skip the redirection page.


chenglou:
Thanks!

Issue #330, opened 2018-03-09
Suggestion: Generate new URL in  “try” only when sharing


jdeisenberg:
Page `https://reasonml.github.io/en/try.html`generates a new URL every time you press a key. These URLs get stored in the browser history, making a search for "reason"-related URLs inconvenient. Would it be possible to generate the compressed code only when clicking the “Share” at the upper right of the window?

PR #537, opened 2019-04-24
Toggle for updating URL

phated:
I added a toggle to the UI that allows someone to disable live updates of the URL (as desired in #330).  It then persists it to localStorage and uses that setting to know whether or not to automatically update the URL on future visits.

The "Share" URL is still generated on-the-fly but the end-user will need to click "Share" to copy it instead of copying from the URL bar.

cc @jdeisenberg


jdeisenberg:
Thank you!

phated:
🤷

PR #538, opened 2019-04-26
fixed two small errors

0xflotus:


docs/if-else.md
```diff
@@ -32,7 +32,7 @@ if (showMenu) {
 
 Which is usually what you'd expect; `displayMenu()` should return unit too.
 
-Here's another way too look at it. This is clearly wrong:
+Here's another way to look at it. This is clearly wrong:
```
jordwalke: This one is not correct. It should remain `to`

PR #539, opened 2019-05-02
Link to mutation page from the let binding page.

benbot:
It's a little confusing to have information on mutable let bindings be nowhere on the "Let Binding" page. 
This change just adds a link and a helpful line.

### DIFF
--- a/docs/let-binding.md
+++ b/docs/let-binding.md
@@ -34,6 +34,8 @@ let message = "bye";
 print_endline(message); /* Prints "bye" */
 ```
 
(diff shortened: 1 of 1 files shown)


docs/let-binding.md
```diff
@@ -34,6 +34,8 @@ let message = "bye";
 print_endline(message); /* Prints "bye" */
 ```
 
+Even though let bindings themselves are immutable, there is a way to make a let binding that acts mutable. Check out [Mutation](https://reasonml.github.io/docs/en/mutation) for more info.
```
yawaramin: I recommend putting this link on line 28, e.g. like:

```
explicitly contains a mutable value, discussed in [Mutation](...)
```

PR #546, opened 2019-06-11
Update JSX to proper syntax

tylerwray:
Updates Syntax of JSX to proper syntax


baransu:
Actually that was valid JSX in Reason. It works totally fine but latests versions or refmt are adding {} to make it consistent.

tylerwray:
Good to know, thanks!

PR #549, opened 2019-06-17
Explain placeholder single-use limit

yawaramin:


docs/pipe-first.md
```diff
@@ -138,3 +138,10 @@ This allows you to pipe into any positional argument. It also works for named ar
 getName(input)
   ->namePerson(~person=personDetails, ~name=_);
 ```
+
+**Note** that we support only one placeholder per function call. If you use more than one, they are considered to be the same argument, repeated. So the following are equivalent:
```
andreypopp: When I read "we support only once placeholder per function call" my initial impression was that multiple occurrences of a placeholder are not allowed. Maybe something like "expressions with placeholder syntax are functions where the placeholder denotes the first and only argument".

yawaramin: Ah, true. My wording can be ambiguous. I'll make a change.

PR #550, opened 2019-07-01
Update variant.md

moismat:
typo/grammar tweak

types also ~~allows~~ allow us not needing to overly design an API ...


chenglou:
Thanks

PR #557, opened 2019-07-24
ignore .idea

Imon-Haque:
.gitignore update to ignore Intellij IDEA project files.


thangngoc89:
This should not be here. If you use IDEA, you can make a global gitignore module

chenglou:
Yep =)

PR #560, opened 2019-08-14
Fix overlay error height

boyswan:
Remove (flex: 1)  to prevent error overlay from hiding the last line in the editor.


davesnx:
Hey, what does this fix? Can you show a prev/after? Thanks!

davesnx:
Thanks!

davesnx:
After this PR

Screenshot 2023-08-29 at 16 08 30

Before 
Screenshot 2023-08-29 at 16 08 51

PR #576, opened 2019-10-17
Document fun pattern match syntax

yawaramin:


chenglou:
Thanks, though trying to de-emphasize this pattern.

PR #580, opened 2019-10-26
fix: link to REPL

enieber:
the @joelmoniz call if your repository is a reference to REPL


enieber:
maybe this pr ca be merge after the PR #465

chenglou:
REPL means real-eval-print-loop. Unrelated, thanks =)

PR #593, opened 2019-12-16
Updates bucklescript playground to 7.0.1

EduardoRFS:
The bucklescript playground was updated to 7.0.1, following the new guide on [Contributing to the BS Playground Bundle](https://github.com/BuckleScript/bucklescript/blob/7.0.1/CONTRIBUTING.md#contributing-to-the-bs-playground-bundle)

**This is important**, the new bucklescript version has a lot of features mainly the new FFI and new record serialization to javascript objects, it being available on the repl would make a lot easier to show the new features.

### Before
image

### After
image


imbsky:
@jordwalke We should merge this.

jordwalke:
Thank you @eduardoRFS

ryyppy:
I think this update broke ReasonReact support... I guess ˋsetupSomeArtifactsˋ doesn't work with the most recent jsoo versions / bucklescript bytecode? We are still in the process of making it easy to bundle up 3rd party modules for the BS playground runtime and have proper instructions for that (will probably be done in Q1 2020).

Reason · Reason lets you write simple, fast and quality type safe code while leveraging both the JavaScript   OCaml ecosystems

jordwalke:
@ryyppy What do you suggest we do in the mean time?
1. Revert this?
2. Hide the Reason React checkmark.
3. Fix forward.

EduardoRFS:
Ok this should be reverted, there seems to have some changes in the API also. I didn't test the options related to it.

I will make the fix and in two or three days we reopen one of this.

PR #594, opened 2020-01-03
Build reason-react using the BS 7.0.1

EduardoRFS:
## Problem

Because #593, the playground support to ReasonReact isn't working, it was using the published version of `reason-react` at npm and it was built using `BS 5.0.1`. 

## How it was solved

I just rebuilt `reason-react` using the new `bs-platform` locally.

This [Dockerfile](https://gist.github.com/EduardoRFS/de84d45955b7e19d7d0ddc694785a38c) can be used to replicate this PR


jordwalke:
@ryyppy What do you think about this fix?

ryyppy:
Just followed the build instructions and it works with current BS master (I avoided the `setupSomeArtifacts.js` script though, to prevent any overrides).

There's still one thing missing:
With the current architecture, refmt.js is shipped separately in `website/static/js/refmt.js`, which is currently a year old. This needs to be updated to `refmt` version `3.5.1` to be in sync with the current BuckleScript version. Will open up a new ticket for that.

ryyppy:
Approved

jordwalke:
Thank you for reviewing @ryyppy and apologies for the delay.

PR #595, opened 2020-01-06
add radity logo

oaksakal:
add radity logo


cristianoc:
That’s rad.

oaksakal:
that was quick. thanks! and yes we are rad-ity.

PR #597, opened 2020-01-07
add codecentric logo to users of ReasonML

marcoemrich:
We've started using ReasonML for a customer project last year - and have been very happy with ReasonML!


cristianoc:
Welcome!

PR #622, opened 2020-03-18
Add Swiss Dev Jobs example image and remove logo

Vrq:


Vrq:
@cristianoc sorry for the trouble - I put the image in the wrong directory in the previous PR.

PR #623, opened 2020-03-18
Update image path to the correct one

Vrq:


Vrq:
@cristianoc forgot about correcting the path, sorry again. I need some sleep now 😴

PR #624, opened 2020-03-23
Add Oxidizing Systems as users of Reason

jerben:
@cristianoc Was so kind to point me here on Discord.


cristianoc:
@jerben the svg did not show in github preview, but I assume it's OK. If not, just send a patch.

PR #638, opened 2020-05-07
Document `fun` syntax

adnelson:
I didn't see the `fun` syntax documented anywhere, so I added a section for it on the pattern matching doc. Arguably it could be better to put it in `function.md`?


kyldvs:
Thanks for the PR, it looks like this got taken care of in the updates to the pattern matching docs: https://reasonml.github.io/docs/en/pattern-matching#fun

I'll go ahead and close this out

PR #640, opened 2020-05-09
Fix ES2020 typo

michael-hamilton:
Assuming this is meant to say ES2020, not 2030


evxn:
No, I believe, It's an exaggeration about some distant future versions of JS.

PR #650, opened 2020-05-24
fixed typos in what-and-why docs

evxn:
eachother OCaml -> each other
langauges -> languages


davesnx:
Thanks, was fixed by some other PR.

PR #652, opened 2020-06-04
Fix typo in record.md

SirFurness:
Changed "add fieds" to "add fields"


kyldvs:
Thanks for the PR! The record docs were rewritten recently so I'll close this out.

PR #659, opened 2020-06-25
Update overview

jordwalke:
This makes the overview section a quick reference sheet, by adding a few bullet points for each of the main sections in the left nav.
This allows us to make the sections in the left nav much more thorough. If we make the left nav sections more thorough, without a more concise quick overview sheet, then it's more difficult for newcomers to quickly get a handle on the most important aspects of the language.

### DIFF
--- a/docs/interop.md
+++ b/docs/interop.md
@@ -2,6 +2,30 @@
 title: Interop
 ---
 
+## Output Overview
+
+Feature                         | Example                              | JavaScript Output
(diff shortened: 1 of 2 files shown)


jordwalke:
Reorganize the JS output into the "Interop" section so that everything about JavaScript output is organized into one convenient place.

rickyvetter:
Looks good to me. Thanks for moving things over to the interop section.

PR #668, opened 2020-07-13
Add recursion docs

kyldvs:
Adds documentation for the different kinds of recursion in Reason.

(I didn't see any existing dedicated page for this, but if one exists I can work with it)

### DIFF
--- a/docs/recursion.md
+++ b/docs/recursion.md
@@ -1,0 +1,118 @@
+---
+title: Recursion
+---
+
+There are three places recursion is commonly used: bindings, types, and modules.
+
+Feature  | Recursive           | Non-recursive
+---------|---------------------|--------------
+Bindings | `let rec fn = ...`  | `let fn = ...`
+Types    | `type t = ...`      | `type nonrec t = ...`
+Modules  | `module rec A ...`  | `module A ...`
+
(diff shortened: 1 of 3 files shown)


docs/recursion.md
```diff
@@ -0,0 +1,118 @@
...
+
+```reason
+let rec infiniteRecursion = () => {
+  infiniteRecursion();
+};
+```
+
+## Mutual Recursion
+
+Mutually recursive functions use the `and` keyword:
```
jordwalke, line 45: It might be worth coming back to this to elaborate (maybe at the end)
`and` makes the bindings happen "simultaneously". If the leading binding is `rec`, it means the set of `and`ed bindings cannot see each other at all. If the leading binding is `rec` then it means all the bindings can see each other, and themselves.

docs/recursion.md
```diff
@@ -0,0 +1,118 @@
...
+}
+and function3 = () => {
+  function1();
+};
+```
+
+## Recursive Types
+
+Types, unlike bindings, are recursive by default. This is because it is
+common for types to be recursive and uncommon for types to be overriden. A
```
jordwalke, line 62: overridden -> shadowed

PR #669, opened 2020-07-14
Document removing this binding from objects

briangorman:
Add documentation for removing the this binding, which can result in an unused variable warning.


davesnx:
Hey, can you expand a bit on context for this? 

`To disable the `this` binding you can use `as`.` doesn't tell much

PR #672, opened 2020-07-23
Update pattern-matching.md

robz:
Update pattern matching docs


docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
 
 ```reason
-let optionBoolToBool = opt =>
-  switch (opt) {
-  | Some(trueOrFalse) => trueOrFalse
-  | None => false
-  };
+type t = A(int) | B(int);
+let data = B(3);
+let A(v) | B(v) = data;
```
kyldvs, line 254: I think this is probable more confusing than helpful, maybe it makes sense to leave this out?

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
+  | _ => 3
+  };
+/* k is 1 */
+```
 
-Using a Reason `switch` for the first time might make you feel like you've been missing out all these years. Careful, for it might ruin other languages for you =).
+### Matching exceptions
 
-If you've tried to refactor a big, nested if-else logic, you might realize it's very hard to get the logic right. On the other hand, pattern matching + tuple conceptually maps to a 2D table, where each cell can be independently filled. This ensures that whenever you need to add a case in the `switch`, you can target that and only that table cell, without messing other cells up.
+If executing an input raises an exception, it can be matched alongside the other values of the input:
```
kyldvs, line 278: Does it make sense to add an example of `try/catch` pattern matching in isolation here too, not just exceptions used in a switch. This gives the impression that this is the only way to handle exceptions to me

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
-let optionBoolToBool = opt =>
-  if (opt == None) {
-    false;
-  } else if (opt == Some(true)) {
-    true;
-  } else {
-    false;
-  };
+let x = 3;
+switch (x) {
```
kyldvs, line 207: Add some newlines

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
+
+The compiler returns a warning if patterns do not cover all possible values of the input. An exception will be thrown at runtime for unmatched inputs.
+
+```reason
+let f = x =>
+  switch (x) {
+  /* Warning: this pattern-matching is not exhaustive. */
+  | 0 => "zero"
+  };
+f(2); /* Exception: Match_failure */
```
kyldvs, line 187: Add some newlines

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
 
-You can put a pattern anywhere you'd put a normal "variable declaration":
+A single block of code can be run for multiple patterns by listing them together. The `|` character can also be used inside patterns to list multiple possibilities.
 
 ```reason
-type leftOrRight =
-  | Left(int)
-  | Right(int);
+let li: list(int) = [1, 2, 3, 4];
+switch (li) {
```
kyldvs, line 167: Add some newlines

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
+### Combining patterns
 
-You can put a pattern anywhere you'd put a normal "variable declaration":
+A single block of code can be run for multiple patterns by listing them together. The `|` character can also be used inside patterns to list multiple possibilities.
 
 ```reason
-type leftOrRight =
-  | Left(int)
-  | Right(int);
+let li: list(int) = [1, 2, 3, 4];
```
kyldvs, line 166: I would use a better variable name like "list"

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
-  | GoodResult(theMessage) => ...
-  | BadResult(errorCode) when isServerError(errorCode) => ...
-  | BadResult(errorCode) => ... /* otherwise */
-  | NoResult => ...
-  };
+type point = {
+  x: int,
+  y: int,
+};
+type t =
```
kyldvs, line 118: Add some newlines

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
+  | A((string, int))
+  | B(r)
+  | C(array(int))
+  | D(list(r));
+
+let x = D([{x: 2, y: 1.2}]);
+switch (x) {
```
kyldvs, line 125: Add some newlines

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
+/* y is now "b_42" */
 ```
 
-Instead, it'd assume you're matching on any string, and binding that to the name `myMessage` in that `switch` case, which is not what you wanted.
+This can be useful when working with the option variant.
 
-### When clauses
+```reason
+let x: option(int) = Some(3);
+let value =
```
kyldvs, line 102: Add some newlines

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
 ```reason
-let myMessage = "Hello";
-switch (greeting) {
-| myMessage => print_endline("Hi to you")
-};
+type t =
+  | A
+  | B(int);
+let x = B(42);
+let y =
```
kyldvs, line 89: Add some newlines

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
 
 ```reason
-let myMessage = "Hello";
-switch (greeting) {
-| myMessage => print_endline("Hi to you")
-};
+type t =
+  | A
+  | B(int);
+let x = B(42);
```
kyldvs, line 88: Add some newlines

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
...
 
 ```reason
-let message =
-  switch (data) {
-  | GoodResult(theMessage) => "Success! " ++ theMessage
-  | BadResult(0 | 1 | 5) => "Something's wrong. It's a server side problem."
-  | BadResult(errorCode) => "Unknown error occurred. Code: " ++ string_of_int(errorCode)
-  | NoResult => "Things look fine"
+let k = 60;
+let y =
```
kyldvs, line 71: Add some newlines

docs/pattern-matching.md
```diff
@@ -1,229 +1,306 @@
 ---
-title: Pattern Matching!
+title: Pattern Matching
 ---
 
-_Make sure you've read up on [Variant](variant.md) first_.
+_Quick overview: [Pattern matching](overview.md#pattern-matching)_
 
-**We're finally here**! Pattern matching is one of _the_ best features of the language. It's like destructuring, but comes with even more help from the type system.
+Pattern matching provides a way to execute code when the shape of some data matches a particular pattern. It's similar to switch-case statements in other languages, but it can be more expressive and includes some extra safeguards.
```
robz, line 7: Add paragraph emphasizing variants

docs/pattern-matching.md
```diff
@@ -1,229 +1,366 @@
...
+```reason
+switch (f()) {
+| 0 => "zero"
+| 1 => "one"
+| k => "another number " ++ string_of_int(k)
+};
+```
 
-You can even switch on string, int and others. You can even have many patterns going to the same result!
+Note that if a variable with the same name already exists in the scope of the switch, then it is not used in the pattern, and is instead shadowed by the pattern's variable. Variables in patterns are declarations of new variables, not references to existing ones.
```
jordwalke, line 71: > , then it is not used in the pattern, 

"then it will be shadowed by a new variable with that same name, throughout the statements to the right hand side of the next `=>`. That original value that was already inside a variable by the same name is not used a matching constraint for the pattern match. Variable names in pattern matches always create new variables in scope.

docs/pattern-matching.md
```diff
@@ -1,229 +1,366 @@
...
-  | NoResult;
+switch (input) {
+| Pattern1 => code1
+| Pattern2 => code2
+| Pattern3 => code3
+};
 ```
 
-While using the `switch` expression on it, you can "destructure" it:
+The switch finds the first pattern that matches the input, and then executes the code that the pattern points to (the code after the `=>`). Code for a pattern can be a single expression, or a block of statements.
```
jordwalke, line 23: Worth mentioning right up front that the `switch` evaluates to the value, and is a valid expression. (That's different than what JS users are accustomed to).

docs/pattern-matching.md
```diff
@@ -1,229 +1,366 @@
...
+  | 1 => "one"
+  | k => "another number " ++ string_of_int(k)
   };
+/* y is "another number 3", k is still 60 */
 ```
 
-Combined with other data structures, pattern matching can produce extremely concise, compiler-verified, performant code:
+### Variants
+
+Patterns can also include variants and data held by variant tags.
```
jordwalke, line 87: and => with:

> Patterns can also include variants with data held by those variant tags.

docs/pattern-matching.md
```diff
@@ -1,229 +1,366 @@
...
+| A("a", _) => 0
+| A(_) => 1
+| B(_, (i, _)) => i
+| C([{x, y}, ..._]) => x + y
+| C([]) => 2
+}
 ```
 
-Now that's just silly =). Let's turn it into pattern-matching:
+Now if someone adds a new variant tag to `t`, the exhaustive warning will require them to consider how it should be handled in this `switch` statement. If the last pattern were just `_`, they would receive no warning.
```
jordwalke, line 244: Nice.

PR #674, opened 2020-07-24
Add docs for basic structures

kyldvs:
Cleans up the documentation for basic structures and consolidates them into one page. Right now this covers lists/arrays/tuples with a link to the record page as well.

We may want to add object in here at some point, or add a link to the dedicated object article like we do for record.


docs/basic-structures.md
```diff
@@ -0,0 +1,212 @@
+---
+title: Basic Structures
+---
+
+_Quick overview: [Basic Structures](overview.md#basic-structures)_
+
+## List
+
+Lists are an immutable, dynamically sized, homogeneous, and ordered collection
```
jordwalke, line 9: collections

docs/basic-structures.md
```diff
@@ -0,0 +1,212 @@
...
+
+#### Is Empty
+
+```reason
+let isEmpty = list == [];
+```
+
+### Pattern Matching
+
+Pattern matching works with lists:
```
jordwalke, line 79: I don't think pattern matching has been introduced at this point yet. Should at least linkify the "pattern matching" word to link to the section so they can learn more.

website/sidebars.json
```diff
@@ -16,11 +16,10 @@
       "regular-expression",
       "boolean",
       "integer-and-float",
-      "tuple",
+      "basic-structures",
```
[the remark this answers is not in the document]
kyldvs: Records are a lot more complex and nuanced compared to arrays and lists (IMO). Especially with the nominal typing and all the "gotchas" that people need to be aware of. Lists and arrays are much more familiar structures

docs/basic-structures.md
kyldvs: Yeah I'll linkify it.

I don't think there needs to be as strict a sequence of docs for the full articles, I think that will detract from the overall quality where circular references like this might be helpful (pattern matching will talk about lists, and this list doc will talk about pattern matching). In any case we can consider these topics introduced in the overview which has a non-circular sequence of topics

PR #676, opened 2020-07-24
Update loops documentation

kyldvs:
These docs are basically the same, just some minor updates. Also removed "Imperative" from the sidebar title so the page is easier to find


docs/loops.md
```diff
@@ -0,0 +1,69 @@
...
+---
+
+_Quick overview: [Loops](overview.md#loops)_
+
+Loops are discouraged in most cases. Instead functional programming patterns
+like `map`, `filter`, or `reduce` can usually be used in their place.
+
+## For Loops
+
+For loops use the `for`, `in`, and `to` keywords. The increment cannot be
```
jordwalke, line 12: Changed from what? Answer: From 1.
To clarify I think it could say "`for` loops always increment or decrement a counter by one on each iteration".

PR #677, opened 2020-07-24
Update destructuring docs

kyldvs:
Just some minor edits. I added a little bit more organization and removed some of the more advanced features.

I'm thinking we might want to consolidate the advanced usages of destructuring into an article under "advanced features" section. I think this could also include the advanced versions involving variants and get them all in one place

### DIFF
--- a/docs/destructuring.md
+++ b/docs/destructuring.md
@@ -2,59 +2,65 @@
 title: Destructuring
 ---
 
-"Destructuring" is a visually concise way of extracting fields from a data structure. You can use destructuring anywhere you'd normally use a variable.
+Destructuring is a concise way to extract data from a structure. Destructuring
+can be used anywhere a variable name might appear.
 
-## Usage
-
-The following binds variables: `ten = 10`, `twenty = 20`
-
-```reason
-let someInts = (10, 20);
-let (ten, twenty) = someInts;
-```
-
-The following binds variables: `name = "Guy"`, `age = 30`
+## Tuples
 
 ```reason
-type person = {name: string, age: int};
-let somePerson = {name: "Guy", age: 30};
-let {name, age} = somePerson;
+let pair = (1, "Hello");
+
+/* Destructuring */
+let (one, hello) = pair;
+
+/* Prints: Hello */
+print_endline(hello);
 ```
 
-When you pull out fields, you can optionally rename the fields. The following binds these instead: `n = "Guy"`, `a = 30`.
+## Records
 
-```reason
-let {name: n, age: a} = somePerson;
-```
-
-Destructuring also allows type annotations.
-
-```reason
-let (ten: int, twenty: int) = someInts;
-let {name: (n: string), age: (a: int)} = somePerson;
-```
-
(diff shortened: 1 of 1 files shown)


docs/destructuring.md
```diff
@@ -2,59 +2,65 @@
...
-The following binds variables: `ten = 10`, `twenty = 20`
+## Tuples
 
 ```reason
-let someInts = (10, 20);
-let (ten, twenty) = someInts;
-```
+let pair = (1, "Hello");
 
-The following binds variables: `name = "Guy"`, `age = 30`
```
jordwalke, line 16: I kind of prefer single line comments for inline code example comments. For some reason it looks a little cleaner/familiar.

docs/destructuring.md
```diff
@@ -2,59 +2,65 @@
...
+let {name, age} = alice;
+
+/* Prints: Alice */
+print_endline(name);
 ```
 
-**Keep reading the section, pattern matching, for a crazier form of destructuring**!
+### Renaming Record Fields
+
+While destructuring it is possible to bind to variables different than the field
```
jordwalke, line 44: A little more accurate:

> While destructuring record fields, it is possible to bind variable names that are different than the record field names.

docs/destructuring.md
```diff
@@ -2,59 +2,65 @@
...
-Destructuring can make your code much more concise without requiring you to name intermediate variables. Do use them! But don't abuse them and make your code overly nested & terse.
+```reason
+/* Without destructuring */
+let hello = (person) => {
+  print_endline("Hello " ++ person.name);
+};
 
-If you're destructuring a record or a variant whose definition isn't in the current file, you need to explicitly annotate it. See [here](record.md#record-needs-an-explicit-definition) and [here](variant.md#variant-needs-an-explicit-definition).
+/* With destructuring */
+let hello = ({name}) => {
```
jordwalke, line 63: Could also add an example of your aliasing inside a function argument. `({name: theName}) => `

docs/destructuring.md
```diff
@@ -2,59 +2,65 @@
...
+## Function Arguments
 
-Destructuring can make your code much more concise without requiring you to name intermediate variables. Do use them! But don't abuse them and make your code overly nested & terse.
+```reason
+/* Without destructuring */
+let hello = (person) => {
+  print_endline("Hello " ++ person.name);
+};
 
-If you're destructuring a record or a variant whose definition isn't in the current file, you need to explicitly annotate it. See [here](record.md#record-needs-an-explicit-definition) and [here](variant.md#variant-needs-an-explicit-definition).
```
jordwalke, line 60: I think this note was pretty helpful. Maybe include it?

docs/destructuring.md
kyldvs: I do too, but the syntax highlighting isn't supported. I'm happy to go back and update them all if support is added

kyldvs: Another consideration is that sketch doesn't support single line comments. If we want examples to be copy/pastable in there then we should stick to `/* */` style comments or get support added in sketch.

PR #678, opened 2020-07-29
Update type documentation

kyldvs:
This updates the type documentation

Consolidated the more on type doc with this one so this version is a bit longer, but it probable makes sense to introduce these topics at once. I've also added a section on opaque types with motivating examples


docs/type.md
```diff
@@ -1,98 +1,197 @@
 ---
-title: Type!
+title: Types
 ---
 
-Types are the highlight of Reason! Here, you get a glimpse of why so many are excited about them.
+Types describe what kind of thing values are. Is this value a string, integer,
+or some complex structure? The type of that value will give a direct answer.
```
jordwalke, line 6: > The type of that value will give a direct answer.
Well, sometimes :D
(Polymorphic types, abstract types).

docs/type.md
```diff
@@ -1,98 +1,197 @@
...
+/* Allowed: addition */
+let nextCount = count + 1;
+
+/* Error: count is not a list */
+let x = List.map(fn, count);
+```
+
+## Annotations
+
+Type annotations can appear almost anywhere. They are not often necessary due to
```
jordwalke, line 46: Worth mentioning that they need to reside within parenthesis, or between commas of lists, tuples, and function arguments/parameters.

jordwalke: Alternatively, you could suggest that they should always be placed with surrounding parenthesis and refmt will remove them where they weren't actually necessary. (If anyone has used Flow they probably would get this).

docs/type.md
```diff
@@ -1,98 +1,197 @@
...
-var myInt2 = 5;
-var myInt3 = 9;
-function addInts(x, y) {
-  return x + y | 0;
-}
-function drawCircle(r) {
-  return "hi";
-}
+
+Type aliases are required in some cases, such as working with
```
jordwalke, line 76: nice.

docs/type.md
```diff
@@ -1,98 +1,197 @@
...
+[variants](variant.md) or [records](record.md).
+
+## Type Arguments
+
+Types can have arguments. This is useful when defining structures that work
```
jordwalke, line 81: Is the word `parameters` okay here?
1. Then we can talk about "parameterized types", whereas I do know know of a shorthand way to say "argumentified".
2. _Technically_, parameters are the defined inputs to a function (or parameterized type), and then arguments are the values supplied to the function when called (or type when instantiated).

"Types can accept type parameters, which are similar to generics in other languages. Parameterized types are useful when defining structures that work..."

Kind of being pedantic, and personally I use them interchangeably, but a few people like the distinction, and no one would complain if we adopted the convention: 
https://en.wikipedia.org/wiki/Parameter_(computer_programming)

docs/type.md
```diff
@@ -1,98 +1,197 @@
...
+
+Arguments are prefixed with a single `'` when defining the type:
+
+```reason
+type list('item) = ...
 ```
-<!--END_DOCUSAURUS_CODE_TABS-->
 
-Note: in the last line, `(~radius as r: int)` is a labeled argument. More on this [here](function.md).
+When using this type as an annotation the argument must be filled in with a
```
jordwalke, line 92: > When using this type as an annotation the argument must be filled in with a concrete type:

I might say "can be" instead of "must be", just because you can also use a `'a` in these places, which means "Figure out the type for me and give it the name `'a`" (not that we have to cover that here).

docs/type.md
```diff
@@ -1,98 +1,197 @@
...
-## Design Decisions
+Use the `Duration` module to create and interact with the opaque type:
 
-Reason's type system is the crystallization of decades of research and engineering. Here are a few highlights:
+```reason
+let oneMinute = Duration.fromSeconds(60);
+let twoMinutes = Duration.add(oneMinute, oneMinute);
+```
+
+Normal integer functions intentionally have errors with `Duration.t` values:
```
jordwalke, line 153: like-stamp

docs/type.md
```diff
@@ -1,98 +1,197 @@
...
+
+### Setup
+
+`Duration.t` is an opaque type. Its concrete type is hidden and it can only be
+interacted with using a limited set of functions:
+
+_(Writing `type t;` is what makes `t` opaque. `t` would not be opaque and would
+still have a concrete type if `type t = int;` were written instead.)_
+
+```reason
```
jordwalke, line 130: I might just explain this in terms of interface files and implementation files since the reader has probably not covered modules yet. Then when modules are covered they can be explained in terms of files ("It's like embedding one Reason file inside of another").

website/sidebars.json
```diff
@@ -12,15 +12,14 @@
       "overview",
       "let-binding",
       "primitives",
-      "type",
       "basic-structures",
+      "type",
```
jordwalke, line 16: Does this make sure the slug stays the same (bookmarks).

kyldvs: Yeah the URL is based entirely on the file name from what I can tell

docs/type.md
kyldvs: That's a lot harder to put into documentation and is more confusing IMO. At this point we also have not covered interface files, they will get covered at the same time as modules pretty much

docs/type.md
kyldvs: > The type of that value gives an answer.

Is that better? Just less explicit. It does give an answer, just maybe it's not a complete answer like you point out :P

PR #680, opened 2020-07-30
Update variant and option documentation

kyldvs:
Updates variant and option docs

- Removed "undefined" from the old title of the option docs since it is JavaScript specific. Instead it is referred to more generally as nullability now


docs/option.md
```diff
@@ -0,0 +1,65 @@
...
+- `Some` represents the presence of a value and accepts that value as an argument
+- `None` represents the absence of a value and acceps no arguments
+
+```reason
+type option('value) =
+  | None
+  | Some('value);
+```
+
+_Note: It is not necessary to define this type, it is already defined._
```
jordwalke, line 28: like-stamp

Maybe "it is already defined and automatically available for use everywhere"

docs/option.md
```diff
@@ -0,0 +1,65 @@
...
+let login = () =>
+  Some({
+    name: "Alice",
+    age: 42,
+  });
+
+let alice = login();
+```
+
+Now when writing functions that deal with users we can make sure that there is
```
jordwalke, line 55: Perhaps: can => must
The type system will require that we handle both cases, as with all variants.

docs/overview.md
```diff
@@ -194,6 +196,8 @@ Using a variant                 | `let x = Two("Hello", "World");`
 
 ## Options
 
+_Details: [Options](option.md)_
```
jordwalke, line 199: I like this format of a brief overview page with links to details. (Kind of wish our docs could expand the sections inline on demand) (or, the entire doc could compress into the equivalent of the overview page at will / by default).

docs/variant.md
```diff
@@ -1,207 +1,200 @@
...
-  | None
-  | Instagram(string)
-  | Facebook(string, int);
+type linkedList =
+  | Node(int, linkedList)
+  | Empty;
 ```
 
-Here, `Instagram` holds a `string`, and `Facebook` holds a `string` and an `int`. Usage:
+The constructors are now like functions that accept their data as arguments.
```
jordwalke, line 47: Might be confusing as some people will try to pass them as first class functions (I _*wish*_!)

docs/variant.md
```diff
@@ -1,207 +1,200 @@
...
-
-The [standard library](/api/index.html) exposes two important variants you'll come to hear a lot.
+## Usage
 
-#### `option`
+The primary way to interact with variants is through
+[pattern matching](pattern-matching.md). When pattern matching over a variant
+the compiler can ensure that all cases are covered exhaustively so that there
+is no undefined behavior. This is also helpful when adding new cases to a
+variant later, the compiler will present a clear list of code that needs
```
jordwalke, line 63: > variant later, because the

docs/variant.md
```diff
@@ -1,207 +1,200 @@
...
+  | Horse => "ride"
+  | Snake => "run away"
+  };
+};
 ```
 
-This is the convention used to simulate a "nullable" (aka `undefined` or `null`) value in other languages. Thanks to this convenience type definition, Reason can default every value to be non-nullable. An `int` will always be an int, never "`int` **or** `null` **or** `undefined`". If you do want to express a "nullable int", you'd use `option(int)`, whose possible values are `None` or `Some(int)`. `switch` forces you to handle both cases; therefore, **a pure Reason program doesn't have null errors**.
-
-#### `list`
+It is possible to interact with a constructor's data while pattern matching:
```
jordwalke, line 77: Maybe we can forgo this example since it is covered extensively in the pattern matching doc?

jordwalke: Or do you think it makes sense to eventually merge pattern matching and variants?

docs/variant.md
```diff
@@ -1,207 +1,200 @@
...
+the arguments names a record is used:
 
 ```reason
-type account =
-  | Facebook(string, int) /* 2 arguments */;
-type account2 =
-  | Instagram((string, int)) /* 1 argument - happens to be a 2-tuple */;
-```
-### Variants Must Have constructors
+type cat = {
```
jordwalke, line 103: Might make sense to actually include more than one field in this since the docs just stated a motivation to avoid "many arguments" (weird to see only one here).

jordwalke: Or you could just omit the first paragraph as people already know this intuitively, and instead setup the example. Often times variants will hold a single predefined type..
`<example>`
But if that record type is only ever used within a single variant, it is often more concise to define the variant and the record type at the same time.

docs/variant.md
```diff
@@ -1,207 +1,200 @@
...
+/* No error */
+let y: Animal.t = Cat;
+```
 
-## Design Decisions
+### Tuples as Arguments
 
-Variants, in their many forms (polymorphic variant, open variant, GADT, etc.), are likely _the_ feature of a type system such as Reason's. The aforementioned `option` variant, for example, obliterates the need for nullable types, a major source of bugs in other languages. Philosophically speaking, a problem is composed of many possible branches/conditions. Mishandling these conditions is the majority of what we call bugs. **A type system doesn't magically eliminate bugs; it points out the unhandled conditions and asks you to cover them**\*. The ability to model "this or that" correctly is crucial.
+There is a subtle difference between constructors that accept two arguments and
+constructors that accept one argument that is a 2-tuple. Pay close attention
```
jordwalke, line 160: like-stamp

docs/variant.md
```diff
@@ -1,207 +1,200 @@
...
-  ...
-} else if (data === 'bird') {
-  ...
-}
+let x = OneArg((1, "one"));
+let y = TwoArgs(2, "two");
 ```
 
-There's a linear amount of branch checking here (`O(n)`). Compare this to using a Reason variant:
+Using a one-argument variant that accepts a 2-tuple is common when dealing with
```
jordwalke, line 174: Not sure this is super helpful, as it's not immediately obvious why this is the case (even though it's true). If you're looking for any lines to cut, this could be a candidate.

docs/variant.md
kyldvs: Yeah I see, I can remove the comparison

docs/variant.md
kyldvs: I wanted to get the bare-minimum for interacting with variants in the actual variant document so that someone reading wouldn't have to go through the pattern matching doc to find this as well. I felt that a simple switch and then a switch that uses the data were a nice medium here. I think a little duplication between the docs are okay

docs/variant.md
kyldvs: Good point, I'll rephrase the setup

docs/variant.md
kyldvs: Yeah I just wanted a little bit of motivation for the section, otherwise I feel like people wouldn't understand when this comes up

docs/variant.md
```diff
@@ -1,207 +1,200 @@
...
+structures that use [type arguments](type.md#type-arguments).
```
kyldvs, line 175: Update to parameter terminology

PR #681, opened 2020-07-30
Add implicit return and if-else to overview

kyldvs:
Implicit return is a pretty important concept that is not explicitly mentioned anywhere, adding to the overview

Also split out the if-else section and removed the dedicated page. If-else is such a common language construct and it's not significantly different in Reason so I don't think it needs its own page. A mention in the overview should be fine


docs/overview.md
```diff
@@ -74,9 +74,34 @@ Comparison                      | `>`, `<`, `>=`, `<=`
 Boolean operations              | `!`, `&&`, <code>&#124;&#124;</code>
 Reference equality              | `===`, `!==`
 Structural equality             | `==`, `!=`
+
+## Implicit Return
+
+There is no `return` keyword in Reason. The last expression in a block or
```
jordwalke, line 80: Did we introduce "expression blocks" at this point in the docs yet? If not, I might include it before this.

docs/overview.md
```diff
@@ -74,9 +74,34 @@ Comparison                      | `>`, `<`, `>=`, `<=`
...
+function definition is the return value for that block.
```
jordwalke, line 81: I don't think that "return values" make sense for blocks, only for functions. Blocks are expressions that evaluate to a value like every other expression, so it's probably more accurate to say that "The last item in an expression block is what that expression block evaluates to". 

Then somehow explain that functions just have a single expression on the right hand side of their `=>`, and if it is a block so be it - and that expression will evaluate like every other expression, with no ability to early return.

docs/overview.md
```diff
@@ -74,9 +74,34 @@ Comparison                      | `>`, `<`, `>=`, `<=`
...
+returned.
+
+## If-Else Expressions
+
+Feature                         | Example
+--------------------------------|----------
 If-Else expressions             | `if (condition) { a; } else { b; }`
 Ternary expressions             | `condition ? a : b;`
 
+- Note: These are expressions and can be assigned to a variable: `let x = if (condtion) { a; } else { b; };`
```
jordwalke, line 103: like-stamp

docs/overview.md
kyldvs: Changed the example to deal with functions so this is less confusing.

docs/overview.md
kyldvs: Removed expression blocks from the example.

PR #682, opened 2020-07-30
Update function

kyldvs:
Update the documentation on functions. Cleans up the cheat sheet into tables, adds some additional detail around advanced features

### DIFF
--- a/docs/function.md
+++ b/docs/function.md
@@ -192,24 +192,47 @@ let c = makeCircle(~radius=1, ());
 
 #### Passing Options to Non-Mandatory Arguments
 
-Sometimes you will have an option that needs to be passed to a function with a
-non-mandatory argument. Passing it normally will not work:
+Sometimes you will want to provide a non-mandatory argument based on whether
+a variable is `Some(data)` or `None`.
 
 ```reason
-let addOne = (~value=?, ()) => { ... };
+let fn = (~data=?, ()) => { ... };
 
-let x = Some(10);
+let x = Some(100);
 
+/* Provide data based on whether x is Some or None */
+switch (x) {
+| Some(data) => fn(~data, ());
+| None => fn();
+};
+```
+
+Passing `x` directly to `~data` will not work because the callsite does not
+expect an option type.
+
+```reason
 /* Error: Incorrect type */
-addOne(~value=x)
+fn(~data=x, ());
 ```
 
-Prefix the variable name with a `?` to explicitly pass the option:
+Reason provides a shorthand syntax for the pattern above. Prefixing a named
+argument's value with a question mark at the call site, will pass whatever is
+inside of the `Some()`, but if the value is `None`, then the named argument
+will not be supplied at all.
 
 ```reason
-/* Good */
-addOne(~value=?x)
+let a = Some(100);
+let b = None;
+
+/* fn is called like `fn(~data=100, ())` */
+fn(~data=?a, ());
+
+/* fn is called like `fn()` */
+fn(~data=?b, ());
 ```
+
+_Note: This shorthand works with non-mandatory arguments that have default
+values too._
 
 #### Referencing Previous Arguments
 
--- a/docs/function.md
+++ b/docs/function.md
@@ -117,9 +117,37 @@ let half = divide(_, 2);
 let five = half(10);
 ```
 
-### Default Argument Values
+### Non-Mandatory Arguments
 
-When defining a function with named arguments default values can be provided:
+When defining a function, named arguments can be marked as non-mandatory.
+This means they do not need to be supplied as arguments by the caller.
+You may simultaneously mark a named argument as non-mandatory, and specify a
+default value (in the event it was not supplied by the caller).
+
+#### Defining Non-Mandatory Arguments
+
+Use `=?` after a named argument to mark it as non-mandatory. Inside the function
+this value will be wrapped in an option based on whether or not the caller
+provided the value.
+
+_Notice: The caller provides an `int`, but inside the function `value` is
+`option(int)`._
+
+```reason
+let addOne = (~value=?, ()) => {
+  switch (value) {
+  | Some(value) => value + 1
+  | None => 1
+  };
+};
+
+addOne(); /* 1 */
+addOne(~value=11, ()); /* 12 */
+```
+
+#### Default Values
+
+Default values can be provided to named arguments using `=value`.
 
 ```reason
 let makeCircle = (~x=0, ~y=0, ~radius=10, ()) => { ... };
@@ -131,15 +159,19 @@ makeCircle();
 makeCircle(~x=10, ~radius=2, ());
 ```
 
-Notice that an extra unit argument was added after the `radius` argument. This
-has to do with the partial application feature. Now that some arguments have
-default values and can be left out the compiler needs some indication of when
-you are "done" calling a function.
+#### Final Unit Argument
 
-There must be a positional argument after all optional arguments so the compiler
-knows whether to partially apply a function or to use the default values.
+In the above examples a final unit argument was added to the function
+definitions. This is related to the partial application feature. When using
+non-mandatory arguments that can be omitted at call sites, the compiler needs
+some indication of when you are "done" calling a function.
 
-Consider if there were no final unit argument, what should this do:
+Specifically, there must be a positional argument after all non-mandatory
+arguments so the compiler can determine whether to partially apply the provided
+arguments or evaluate the function using default values.
+
+Consider if there were no final unit argument to `makeCircle`, what
+should this do:
 
 ```reason
 makeCircle(~radius=1);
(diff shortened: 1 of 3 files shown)


docs/function.md
```diff
@@ -110,231 +108,144 @@ let eleven = addFive(6);
...
+/* Creates a circle at (0, 0) with radius 1 */
+let c = makeCircle(~radius=1, ());
 ```
 
-### Explicitly Passed Optional
+### Optional Arguments
 
-Sometimes, you might want to forward a value to a function without knowing whether the value is `None` or `Some(a)`. Naively, you'd do:
+Similar to default argument values, named arguments can be marked as optional.
+This is just syntax around making the default value `None`.
```
jordwalke, line 164: This is not exactly correct:

https://sketch.sh/s/yJfd9kE6DmzSgxP4dx9LBJ/

- A question optional means "there is no default for this argument, and inside of the function, the variable will be either `None` (if no argument was passed) or `Some(arg)` if argument `arg` was passed.
- A default value of `~arg=None`, means that the argument supplied must be an optional, and if no `~arg` argument is supplied, the `arg` identifier inside the function will receive the `None` value by default. If an argument `~arg` _was_ supplied then the value of `arg` inside of the function will be the value passed to the function when called - which will be an optional value.

docs/function.md
```diff
@@ -110,231 +108,144 @@ let eleven = addFive(6);
...
-let rec neverTerminate = () => neverTerminate();
+next(); /* 1 */
+next(~value=11, ()); /* 12 */
 ```
 
-### Mutually Recursive Functions
-
-Mutually recursive functions start like a single recursive function using the
-`rec` keyword, and then are chained together with `and`:
+Sometimes you will have an option that needs to be passed to a function with an
```
jordwalke, line 178: I realized that the terminology of "optional argument" is confusing because it has the word "option" in it. That might be the main source of confusion I see from people with respect to named arguments with defaults. That immediately makes people think that the type they pass to the function is of type `option<t>`, when it's not the case.

What could fix that confusion by using the terminology "defaulted argument". A function that accepts a `~name` argument that is defaulted, _still_ accepts a `string`. It just happens to be defaulted to something if it wasn't supplied by the caller.

Then this section could be much less confusing if described like:
> Sometimes you will have a function that specifies a named argument with a defaulted value. The caller can supply that argument, or omit it. It is common to decide whether or not to supply that argument based on whether or not some option variable is `None` vs. `Some(data)`.
> ```reason
> switch(myVariable) {
> | Some(data) => callFunction(~data=data, 100)
> | None => callFunction(100)
> }
> ```
> Since this is a common pattern, Reason provides a shorthand syntax for the pattern above:
> Prefixing the named argument value with a question mark, will cause that named argument to
> be supplied with the contents of `Some(contents)` if the value is indeed `Some(..)`, otherwise
> the named argument will not be supplied at all.
> ```reason
> callFunction(~data=?data, 100)
> }
> ```

docs/function.md
```diff
@@ -110,231 +108,144 @@ let eleven = addFive(6);
...
 
-If we _do_ supply the unit, OCaml knows we deliberately omit the `radius` parameter, and the function is executed.
-```reason
-let circle = drawCircle(~color, ());
+/* Creates a circle at (0, 0) with radius 1 */
+let c = makeCircle(~radius=1, ());
 ```
 
-### Explicitly Passed Optional
+### Optional Arguments
```
jordwalke, line 161: nonmandatory

docs/function.md
```diff
@@ -110,231 +108,144 @@ let eleven = addFive(6);
...
 ```
 
-When given in this syntax, `radius` is **wrapped** in the standard library's `option` type, defaulting to `None`. If provided, it'll be wrapped with a `Some`. So `radius`'s type value is either `None` or `Some(int)` here.
+Notice that an extra unit argument was added after the `radius` argument. This
+has to do with the partial application feature. Now that some arguments have
+default values and can be left out the compiler needs some indication of when
+you are "done" calling a function.
 
-**Note**: `None | Some(foo)` is a data structure type called variant, described [earlier](variant.md). This particular variant type is provided by the standard library. It's called `option`. Its definition: `type option('a) = None | Some('a)`.
+There must be a positional argument after all optional arguments so the compiler
```
jordwalke, line 139: nonmandatory

docs/function.md
```diff
@@ -110,231 +108,144 @@ let eleven = addFive(6);
...
 
 ```reason
-let add = (x) => (y) => x + y;
+let divide = (a, b) => a / b;
+let half = divide(_, 2);
+let five = half(10);
 ```
 
-OCaml optimizes this to [avoid the unnecessary function allocation](https://reasonml.github.io/en/try.html?reason=DYUwLgBAhgJjEF4IAoAeBKRA+FBPTCOqEA1BLgNwBQVA9AFQTAD2zA1tJGABYgTMBXMAAchAQmhwAYgEsAbnxkBnaBAD6SmQDsA5qDUQAZgK0BjMDOZaIpqMGAT6tKqEiwYshYkkxkAVnRqF3AITWIkd08QZABGQKA) (2 functions here, naively speaking) whenever it can! This way, we get
+### Default Argument Values
```
jordwalke, line 120: How about:

> ### Nonmandatory Arguments
> When defining a function, named parameters can be marked as non-mandatory, which means they do not _need_ to be supplied as arguments by the caller. You may simultaneously mark a named parameter as non-mandatory, and specify a default value (in the event it was not supplied by the caller) by appending `=defaultValue` to the named paramter.

docs/function.md
```diff
@@ -110,231 +108,144 @@ let eleven = addFive(6);
...
-  | Some(r_) => startAt(r_, r_)
-  }
-};
+/* Position (10, 0) with radius 2 */
+makeCircle(~x=10, ~radius=2, ());
 ```
 
-When given in this syntax, `radius` is **wrapped** in the standard library's `option` type, defaulting to `None`. If provided, it'll be wrapped with a `Some`. So `radius`'s type value is either `None` or `Some(int)` here.
+Notice that an extra unit argument was added after the `radius` argument. This
+has to do with the partial application feature. Now that some arguments have
```
jordwalke, line 135: > Now that some parameters are marked as non-mandatory, and can be omitted at the call sites, the compiler..

docs/function.md
```diff
@@ -110,231 +108,144 @@ let eleven = addFive(6);
...
-let circle = drawCircle(~color, ());
+/* Creates a circle at (0, 0) with radius 1 */
+let c = makeCircle(~radius=1, ());
 ```
 
-### Explicitly Passed Optional
+### Optional Arguments
 
-Sometimes, you might want to forward a value to a function without knowing whether the value is `None` or `Some(a)`. Naively, you'd do:
+Similar to default argument values, named arguments can be marked as optional.
```
jordwalke, line 163: Instead we could say:
> Sometimes inside of a function definition, you may not wish to provide a default value for a non-mandatory parameter. The syntax for doing so, is to provide a defaulted value of `?`.
```reason
let myFun = (~namedArg=?, ()) => ...
```

When doing so, and as a convenience to the programmer, the named parameter will be converted into an `option<'t>` value inside of the function body, and the function may be pattern matched. If the named parameter is `None`, then it means the named argument was not supplied by the caller. If the named parameter value is `Some(v)` then it means the named argument _was_ supplied by the caller, and the value `v` was supplied.

docs/function.md
jordwalke: I think my comment above might provide an alternative explanation that clarifies this.

docs/function.md
```diff
@@ -110,231 +108,144 @@ let eleven = addFive(6);
...
-
-/* explicit optional application */
-add(~first=?Some(1), ~second=?Some(2));
-/* with punning */
-add(~first?, ~second?);
-```
+### Referencing Previous Arguments
 
-#### With Type Annotation
+When defining default values previous arguments may be used:
```
jordwalke, line 196: like-stamp

kyldvs:
Thanks for the feedback, I've gone through and updated with the non-mandatory terminology. I think it's a lot more clear now.

docs/function.md
```diff
@@ -110,231 +108,185 @@ let eleven = addFive(6);
 let twelve = addFive(7);
 ```
 
-Actually, the above `add` is nothing but syntactic sugar for this:
+If you want to partially call a function with an argument that is not the first
+argument use a `_` to skip arguments:
 
 ```reason
-let add = (x) => (y) => x + y;
```
jordwalke, line 116: This was actually an interesting bit. I'll think about including it in future iterations somehow.

docs/function.md
```diff
@@ -110,231 +108,185 @@ let eleven = addFive(6);
...
+let divide = (a, b) => a / b;
+let half = divide(_, 2);
+let five = half(10);
 ```
 
-OCaml optimizes this to [avoid the unnecessary function allocation](https://reasonml.github.io/en/try.html?reason=DYUwLgBAhgJjEF4IAoAeBKRA+FBPTCOqEA1BLgNwBQVA9AFQTAD2zA1tJGABYgTMBXMAAchAQmhwAYgEsAbnxkBnaBAD6SmQDsA5qDUQAZgK0BjMDOZaIpqMGAT6tKqEiwYshYkkxkAVnRqF3AITWIkd08QZABGQKA) (2 functions here, naively speaking) whenever it can! This way, we get
+### Non-Mandatory Arguments
+
+When defining a function, named arguments can be marked as non-mandatory.
```
jordwalke, line 122: Looks great. Not a mention of "option" anywhere (confusing).
In future iterations we could consider the term "opt-in arguments".

PR #684, opened 2020-08-07
Update modules docs

kyldvs:
This updates the module documentation and cleans up some of the content.

### DIFF
--- a/docs/module.md
+++ b/docs/module.md
@@ -107,28 +107,24 @@ It is covered [later](#including-modules).
 ## Including Modules
 
 The `include` keyword copies the contents of a module into another module. This
-can provide a sense of inheritance by always including a base module.
+can provide a sense of inheritance by always including a default set of
+functions.
 
 ```reason
-module BaseItem = {
-  let name = "Unknown";
-  let cost = 0;
-  let quantity = 0;
-
-  let getTotalCost = () => cost * quantity;
+module BaseItem = {  
+  let isEnabled = (user) => user.isLoggedIn;
+  let run = () => "Do Nothing.";
 };
 
-module Apples = {
-  include BaseItem;
-
-  /* These overwrite the values from BaseItem */
-  let name = "Apple";
-  let cost = 50;
-  let quantity = 7;
+module Hello = {
+  include BaseItem;  
+  let run = () => "Hello World!";
 };
 
-/* We do not have to re-implement this function for each item. */
-Apples.getTotalCost();
+/* Can safely use the interface of BaseItem even if isEnabled does not change */
+if (Hello.isEnabled(User.getCurrentUser())) {
+  print_endline(Hello.run());
+};
 ```
 
 ## Module Types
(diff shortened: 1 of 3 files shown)


docs/module.md
```diff
@@ -1,341 +1,246 @@
...
-};
+## Files are Modules
 
-let message = MyModule.NestedModule.message;
-```
+Every Reason (`.re`) file creates a module for the contents of that file. This
+is commonly how modules are created and code splitting occurs.
 
-### `open`ing a module
+A file named `Duration.re` (or `duration.re`) will result in a module named
```
jordwalke, line 37: good callout on the lowercase

docs/module.md
```diff
@@ -1,341 +1,246 @@
...
+  let name = "Apple";
+  let cost = 50;
+  let quantity = 7;
+};
 
-A module's type is called a "signature", and can be written explicitly. If a
-module is like a `.re` (implementation) file, then a module's signature is like
-a `.rei` (interface) file.
+/* We do not have to re-implement this function for each item. */
+Apples.getTotalCost();
```
jordwalke, line 131: It was nice that you showed how new bindings overwrite the previous ones, which is sometimes what you want, however, this comment here could be misleading "We do not have to re-implement this function for each item. " because in this case, the old `getTotalCost()` references the old values not the overridden ones. There's a good chance the user _does_ have to re-implement this function _in this case_. Maybe consider rewording to:
> `Apples` inherits `getTotalCost`, but `getTotalCost` still refers to the _previous_ versions of `name`, `cost`, and `quantity`.

docs/module.md
```diff
@@ -1,341 +1,246 @@
...
 };
 ```
 
-**Note**: `BaseComponent` is a module **type**, not an actual module itself!
+### Accessing Implicit Module Type
 
-If you do not have a defined module type, you can extract it from an actual module
-using `include (module type of ActualModuleName)`. For example, we can extend the
-`List` module from the OCaml standard library, which does not define a module
+Sometimes you may need access to the type of a module, but that module will not
```
jordwalke, line 181: We may wish to include a word of warning about this feature. I have heard that there are some pretty serious problems with it and might eventually be re-worked in the language. (Or we could even omit it so that newcomers don't accidentally reach for it - allowing experienced devs to use it as an escape hatch).

docs/module.md
```diff
@@ -1,341 +1,246 @@
...
+module type PrintableDuration = {
+  include (module type of Duration);
+  include Printable;
 };
 ```
 
-### Every `.rei` file is a signature
+## Interface Files
+
+Interface files (`.rei` files) are similar to module types. They create a module
```
jordwalke, line 194: Might even make a stronger claim - that `.rei` files _are_ module types applied to an entire file and everything you learn/know about module types can directly be used inside of a `.rei` file.

docs/module.md
kyldvs: Nice catch, I did it backwards! I'll update this example to something that makes more sense.

jordwalke:
Otherwise this should be ready to go.

PR #694, opened 2020-08-30
Fix typos

twitchard:
Noticed "missiles" was spelled wrong. While I was in there, made some other typographical suggestions.


davesnx:
Thanks @twitchard

PR #696, opened 2020-09-07
Update what and why

kyldvs:
Simplifies the what & why documentation and adds links to others' experiences using Reason


davesnx:
That's the simplification that we need thanks @kyldvs. rescript diverge makes the explanation much simpler.

kyldvs:
Thanks, I think this is one of the last pages to simplify! @jordwalke does this look good to you? I'll merge with your approval

PR #698, opened 2020-09-11
Update what-and-why.md

ryyppy:
The "what & why" were outdated... i removed the BuckleScript parts and added ReScript to the alternative section


ryyppy:
whoops nevermind, currently in the works in #698

PR #707, opened 2020-10-13
Updated "bsb -init" example

spookylukey:
Old example gives:
```
$ bsb -init my-new-project -theme basic-reason
Making directory my-new-project                      
Available themes:                
basic                                                                                              
generator                                
minimal                                      
node                       
react-hooks                                                         
react-starter                                                         
tea                                                       
Error: theme basic-reason not found
```

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -19,7 +19,7 @@ yarn global add bs-platform
 The global installation comes with a simple project generator. Try:
 
 ```sh
-bsb -init my-new-project -theme basic-reason
+bsb -init my-new-project -theme basic
 ```
 
 To compile & run the project you just created:


spookylukey:
There are actually more issues. The text below says:

> By default, the basic-reason theme configures BuckleScript to output the generated JS files alongside the Reason files they were produced from. There's a reason why this is helpful. If you prefer to keep the generated files somewhere else, edit bsconfig.json to set in-source to false; the JS files will then be output to the lib/js directory instead.

So `basic-reason` no longer exists, but `basic` does not behave as described here, so it isn't a simple drop-in replacement. Don't know what should be done about this.

kyldvs:
This article needs to be completely rewritten to address multiple environments, new editor plugins, ReScript, etc. It's on my radar!

spookylukey:
I'll close this this PR to reduce the noise and leave you to it, thanks! I thought it was a simple fix when I opened it.

PR #710, opened 2020-11-23
recursion: fix multiple definition of the type name t

raptazure:
Example of [Recursive Types](https://reasonml.github.io/docs/en/recursion#recursive-types) gives the following error:
```
Error: Multiple definition of the type name t.
       Names must be unique in a given structure or signature.
```
Fixed it by renaming the second type name.

### DIFF
--- a/docs/recursion.md
+++ b/docs/recursion.md
@@ -76,10 +76,10 @@ Opt-out of recursive types using the `nonrec` keyword:
 
 ```reason
 type t = string;
-type nonrec t = list(t);
+type nonrec t' = list(t);
 
-/* t is now list(string) */
-let x: t = ["hello", "world"];
+/* t' is now list(string) */
+let x: t' = ["hello", "world"];
 ```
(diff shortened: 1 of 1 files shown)


kyldvs:
Thanks for the PR, for 1. your change kind of defeats the purpose of the example, with different names the `nonrec` keyword is no longer necessary and could be removed. It's necessary for the types to have the same name in order for this keyword to make sense.

I do not believe the example is an error, but could be one, or perhaps a warning based on your compilation settings: https://sketch.sh/s/p5Bc8Y3L8qoufPo7jxhlGe/

raptazure:
Oh, thanks for responding.
I notice that this example works fine with the repl but gives a compilation error when running `yarn build`.  Same with OCaml (`utop` works fine while `ocamlc` gives an error). The following code fixes it:

```reason
type t = string;
module M = {
  type nonrec t = list(t);
};
```

PR #706, opened 2020-10-05
fix: missing yarn install step

JulianLeviston:
First up, massive massive props for making it relatively easy to edit your documentation and suggest PRs on it. Well done there!

I was going through the installation instruction today (new reason user) and I noticed this was missing.
For seasoned JS devs used to yarn/npm, it's obvious what to do, but I thought it wouldn't hurt to add it for anyone else.

PR #717, opened 2020-12-15
Update installation.md

codemilli:
I should install npm packages with `yarn` command before `yarn build`


codemilli:
Duplicated https://github.com/reasonml/reasonml.github.io/pull/706

PR #768, opened 2023-11-26
Fix misleading example of open object types

YawarRaza7349:
The given example using an open object type isn't how one would actually use it. The point of an open object type is so one can write code that works across multiple concrete object types that share some, but not all, of their fields. I made it so that the example is closer to how one would actually use the language feature to achieve this purpose.

Also, I think there's a lot of room for improvement for the Object page as a whole. What do other people think?

### DIFF
--- a/docs/object.md
+++ b/docs/object.md
@@ -26,7 +26,7 @@ type car('a) = {
 } as 'a;
 ```
 
-Two dots, also called an elision, indicate that this is an "open" object type, and therefore can also contain other values and methods. An open object is also polymorphic and therefore requires a parameter.
+Two dots, also called an elision, indicate that this is an "open" object type, and therefore can also contain other values and methods. An open object is also polymorphic and therefore requires a parameter. This parameter refers to the complete type of the object, including those other values and methods.
 
 ### Creation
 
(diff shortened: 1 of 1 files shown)


docs/object.md
```diff
@@ -26,7 +26,7 @@ type car('a) = {
 } as 'a;
 ```
 
-Two dots, also called an elision, indicate that this is an "open" object type, and therefore can also contain other values and methods. An open object is also polymorphic and therefore requires a parameter.
+Two dots, also called an elision, indicate that this is an "open" object type, and therefore can also contain other values and methods. An open object is also polymorphic and therefore requires a parameter. This parameter refers to the complete type of the object, including those other values and methods.
```
davesnx, line 30: Agree here with you that this isn't right, since open object is the form of structural typing.

I'm not sure if this entire paragraph make sense after your addition thought. Since you refer to the `as 'a` without being very explicit.

YawarRaza7349: Just checking: By "Agree with you that this isn't right", are you referring to the old code example, rather than the part that is quoted here?

I don't see how the sentence I added would cause the paragraph to make less sense than it did previously. It already referred to the `'a` when it said "and therefore requires a parameter".

